PR #5, opened 2024-03-28
feat: SQLite integration

Kayra1:
# Description

This change brings CRUD and DB connection and disconnection functions for SQLite as well as simple validators for CSR and Certificates that will go into this DB.

# Checklist:


Kayra1:
I'm not sure about this linter. All of these missing errchecks are intended.

internal/certdb/certdb.go
```diff
@@ -0,0 +1,115 @@
...
+	conn  *sql.DB
+}
+
+type CertificateRequest struct {
+	ID          int
+	CSR         string
+	Certificate *string
+}
+
+func (db *CertificateRequests) RetrieveAll() ([]*CertificateRequest, error) {
```
ghislainbourgeois, line 29: I understand the reason for using a pointer receiver for this function, but not sure about the return being a pointer. The struct only contains basic types so should be relatively cheap to copy.

ghislainbourgeois: The same comment apply for all the return values that are pointers.

internal/certdb/certdb.go
```diff
@@ -0,0 +1,115 @@
...
+		var csr CertificateRequest
+		if err := rows.Scan(&csr.ID, &csr.CSR, &csr.Certificate); err != nil {
+			return nil, err
+		}
+		allCsrs = append(allCsrs, &csr)
+	}
+	return allCsrs, nil
+}
+
+func (db *CertificateRequests) Retrieve(csr *string) (*CertificateRequest, error) {
```
ghislainbourgeois, line 47: Is there a reason I am missing why the parameter is a pointer? Usually basic types are passed by value in Go, as they are cheap to copy.

ghislainbourgeois: This comment is valid for all the function parameters of `*string`.

internal/certdb/certdb.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	db.table = tableName
+	db.conn = conn
+	if _, err := db.conn.Exec(fmt.Sprintf(queryCreateTable, db.table)); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (db *CertificateRequests) Disconnect() error {
```
ghislainbourgeois, line 107: It would be a bit more idiomatic to name this function `Close()`, but that is really minor.

internal/certdb/certdb_test.go
```diff
@@ -0,0 +1,117 @@
...
+	"log"
+	"strings"
+	"testing"
+
+	"github.com/canonical/gocert/internal/certdb"
+)
+
+func TestConnect(t *testing.T) {
+	db := new(certdb.CertificateRequests)
+	defer db.Disconnect()
```
ghislainbourgeois, line 13: The linting is complaining about the error not being checked here. A way to do it in a defer clause is to pass an anonymous function:
```suggestion
	defer func() {
	    if err := db.Disconnect(); err != nil {
	        log.Errorf("Got error: %v\n", err)
	    }
	}()
```

internal/certdb/validation.go
```diff
@@ -0,0 +1,48 @@
+package certdb
+
+import (
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/pem"
+	"errors"
+)
+
+func ValidateCertificateRequest(csrString *string) error {
```
ghislainbourgeois, line 10: Receiving pointers to string here also.

ghislainbourgeois (changes requested):
Great start! I have some concerns around the pointers vs value semantics in the design.

internal/certdb/certdb.go
```diff
@@ -0,0 +1,115 @@
...
+
+type CertificateRequests struct {
+	table string
+	conn  *sql.DB
+}
+
+type CertificateRequest struct {
+	ID          int
+	CSR         string
+	Certificate *string
```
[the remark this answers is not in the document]
Kayra1: This field starts off as null in the DB

internal/certdb/certdb.go
Kayra1: I guess I'm used to using pointers by default. Updated.

internal/certdb/certdb.go
Kayra1: Pointers are all gone.

internal/certdb/certdb_test.go
Kayra1: I know, but is it worth it to add this to everywhere where I'm not testing the disconnect? It will just grow the code in size.

internal/certdb/certdb.go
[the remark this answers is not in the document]
ghislainbourgeois: I think for the public API we could simply return an empty string. It is still clear that there is no certificate in that case.

internal/certdb/certdb_test.go
ghislainbourgeois: You can also add a comment to disable the linting check, I would not personally be against it in tests.

internal/certdb/validation.go
```diff
@@ -0,0 +1,48 @@
+package certdb
+
+import (
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/pem"
+	"errors"
+)
+
+func ValidateCertificateRequest(csrString string) error {
```
ghislainbourgeois, line 10: Please add docstrings on the public functions in this file too.

internal/certdb/validation_test.go
```diff
@@ -0,0 +1,160 @@
...
+		t.Fatalf("Expected error not found:\nReceived: %s", err)
+	}
+	var InvalidCSR = strings.ReplaceAll(ValidCSR1, "/", "p")
+	err = certdb.ValidateCertificateRequest(InvalidCSR)
+	if err == nil {
+		t.Fatalf("Expected CSR to fail validation")
+	}
+}
+
+// Fuzz test
```
ghislainbourgeois, line 124: There are some comments here that I am not sure are needed, are they required?

internal/certdb/certdb.go
```diff
@@ -0,0 +1,115 @@
...
+		var csr CertificateRequest
+		if err := rows.Scan(&csr.ID, &csr.CSR, &csr.Certificate); err != nil {
+			return nil, err
+		}
+		allCsrs = append(allCsrs, csr)
+	}
+	return allCsrs, nil
+}
+
+func (db *CertificateRequests) Retrieve(csr string) (*CertificateRequest, error) {
```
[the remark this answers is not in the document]
Kayra1: What should I return if there isn't a string with this name? A fake object with an empty string as a csr and a -1 as an id? I feel like here it makes sense to return a pointer.

internal/certdb/certdb.go
```diff
@@ -0,0 +1,115 @@
...
+const queryGetAllCSRs = "SELECT rowid, * FROM %s"
+const queryGetCSR = "SELECT rowid, * FROM %s WHERE CSR=?"
+const queryCreateCSR = "INSERT INTO %s (CSR) VALUES (?)"
+const queryUpdateCSR = "UPDATE %s SET Certificate=? WHERE CSR=?"
+const queryDeleteCSR = "DELETE FROM %s WHERE CSR=?"
+
+type CertificateRequests struct {
+	table string
+	conn  *sql.DB
+}
```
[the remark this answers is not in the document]
Kayra1: Done

internal/certdb/certdb_test.go
```diff
@@ -0,0 +1,117 @@
...
+	if err := db.Connect(":memory:", "CertificateReqs"); err != nil {
+		t.Fatalf("Can't connect to SQLite: %s", err)
```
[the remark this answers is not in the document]
Kayra1: I feel like there's no point in continuing a test if one fails, since we'll have to address it regardless and the tests complete in less than a second.

internal/certdb/certdb.go
Kayra1: Done

internal/certdb/certdb.go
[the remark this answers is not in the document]
Kayra1: I actually started off with that, that makes sense. Changed. I think it also makes sense to put a const string to indicate if this csr was rejected in the future.

internal/certdb/certdb_test.go
Kayra1: Awesome. Done.

internal/certdb/certdb.go
```diff
@@ -0,0 +1,115 @@
...
+
+func (db *CertificateRequests) Delete(csr string) error {
+	_, err := db.conn.Exec(fmt.Sprintf(queryDeleteCSR, db.table), csr)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (db *CertificateRequests) Connect(databasePath string, tableName string) error {
```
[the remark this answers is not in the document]
Kayra1: I like this, I wasn't sure factories were encouraged in Go.

internal/certdb/certdb_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+}
+
+func Example() {
+	db := new(certdb.CertificateRequests)
+	if err := db.Connect("./certs.db", "CertificateReq"); err != nil {
+		log.Fatalln(err)
+	}
+	defer db.Disconnect()
+}
```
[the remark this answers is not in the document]
Kayra1: Added some more stuff here

internal/certdb/validation.go
Kayra1: Done

internal/certdb/certdb.go
```diff
@@ -0,0 +1,115 @@
+package certdb
```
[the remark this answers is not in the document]
Kayra1: Done

internal/certdb/validation_test.go
```diff
@@ -0,0 +1,160 @@
...
+	if err := certdb.ValidateCertificateRequest(ValidCSR1); err != nil {
+		t.Fatalf("Couldn't verify valid CSR: %s", err)
+	}
+	if err := certdb.ValidateCertificateRequest(ValidCSR2); err != nil {
+		t.Fatalf("Couldn't verify valid CSR: %s", err)
+	}
+	if err := certdb.ValidateCertificateRequest(ValidCSR3); err != nil {
+		t.Fatalf("Couldn't verify valid CSR: %s", err)
+	}
+}
```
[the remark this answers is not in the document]
Kayra1: I saw this but thought they were overkill, but it looks fun. Done

internal/certdb/validation_test.go
```diff
@@ -0,0 +1,160 @@
...
+
+func TestCSRValidationFail(t *testing.T) {
```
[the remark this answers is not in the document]
Kayra1: With subtests this looks and feels much better to do

internal/certdb/validation_test.go
Kayra1: I was planning on doing some extra stuff, but they're overkill especially since they're just gonna test golang stdlib instead of the validations. removed comments.

internal/certdb/certdb.go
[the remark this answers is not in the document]
Kayra1: I don't really like the above, but it seems to be standard in Go, so I've changed this to be like how I said above.

internal/certdb/validation.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+	return nil
+}
+
+// ValidateCertificate validates the given Cert string to the following:
+// The given CSR must pass the validation provided by ValidateCertificateRequest
+// The cert string must be a valid PEM string, and should be of type CERTIFICATE
+// The PEM string should be able to be parsed into a x509 Certificate
+// The given cert and CSR must share the same public key
+func ValidateCertificate(certString string, csrString string) error {
```
saltiyazan, line 33: I can imagine a function that verifies the format of the certificate like done in `ValidateCertificateRequest` being useful on its own.
Should this be two separate functions?

internal/certdb/validation.go
gruyaume, line 1: I understand the value of these helpers but do they belong under `internal/certdb`. Should we aim for what's here to be purely DB related activities and nothing else?

internal/certdb/validation.go
```diff
@@ -0,0 +1,56 @@
...
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/pem"
+	"errors"
+)
+
+// ValidateCertificateRequest validates the given CSR string to the following:
+// The string must be a valid PEM string, and should be of type CERTIFICATE REQUEST
+// The PEM string should be able to be parsed into a x509 Certificate Request
+func ValidateCertificateRequest(csrString string) error {
```
gruyaume, line 13: Do we need the type in the variable name? Ain't `csr` sufficient?

internal/certdb/certdb.go
```diff
@@ -0,0 +1,136 @@
...
+	if err := db.conn.Close(); err != nil {
+		return err
+	}
+	return nil
+}
+
+// NewCertificateRequestsRepository connects to a given table in a given database,
+// stores the connection information and returns an object containing the information.
+// The database path must be a valid file path or ":memory:".
+// The table will be created if it doesn't exist in the format expected by the package.
```
gruyaume, line 123: Is there a way to enforce docstring in public functions? We are already using `golangci-lint`, maybe we can simply enable one of its linters. In any case, this can be enabled in a PR of its own.

Reference:
- https://golangci-lint.run/usage/linters/

internal/certdb/validation.go
```diff
@@ -0,0 +1,56 @@
...
+// The given CSR must pass the validation provided by ValidateCertificateRequest
+// The cert string must be a valid PEM string, and should be of type CERTIFICATE
+// The PEM string should be able to be parsed into a x509 Certificate
+// The given cert and CSR must share the same public key
+func ValidateCertificate(certString string, csrString string) error {
+	if err := ValidateCertificateRequest(csrString); err != nil {
+		return err
+	}
+	csrBlock, _ := pem.Decode([]byte(csrString))
+	csr, _ := x509.ParseCertificateRequest(csrBlock.Bytes)
```
gruyaume, line 38: Shouldn't we check for the error? This applies at a couple of places here

internal/certdb/certdb.go
```diff
@@ -0,0 +1,136 @@
...
+
+// Retrieve gets a given CSR from the repository.
+// It returns the row id and matching certificate alongside the CSR in a CertificateRequest object.
+func (db *CertificateRequestsRepository) Retrieve(csr string) (CertificateRequest, error) {
+	var newCSR CertificateRequest
+	row := db.conn.QueryRow(fmt.Sprintf(queryGetCSR, db.table), csr)
+	if err := row.Scan(&newCSR.ID, &newCSR.CSR, &newCSR.Certificate); err != nil {
+		newCSR.ID = -1
+		newCSR.Certificate = ""
+		newCSR.CSR = ""
```
ghislainbourgeois, line 60: I do not think this is necessary, the struct fields will be initialized to their zero values by default, so it will contain 0, "", "" by default. In Go, it is idiomatic to treat the returned value as invalid if an error is returned.

ghislainbourgeois (changes requested):
This looks good to me, and I agree with the other comments on `validation.go`.

internal/certdb/validation.go
Kayra1: I think it makes sense to have the validation on the db level. If we have multiple input sources ending up at the db, it will help us avoid repeating the same error checks everywhere and risking having different rules for certs coming from an upload vs another program for example.

internal/certdb/validation.go
Kayra1: We don't check it here because `ValidateCertificateRequest` already catches these errors and returns before we can get here. If there was an uncaught error, golangci-lint will not pass with an (errcheck) error.

internal/certdb/certdb.go
Kayra1: That's nice. deleted it.

internal/certdb/certdb.go
Kayra1: I read the disabled ones but none of them stuck out for me. We could eventually write one ourselves.

internal/certdb/validation.go
Kayra1: I avoided doing that because I didn't want to give the impression that these are helper functions to be used anywhere we need to look at certificates, but there is really no downside to doing that if that's preferred. I've changed it.

internal/certdb/validation.go
Kayra1: yes, I reworked these a bit now.

ghislainbourgeois (approved):
LGTM

gruyaume (approved):

PR #6, opened 2024-04-04
feat: HTTPS server skeleton

Kayra1:
# Description

This change introduces some helper functions and initialization code to have a barebones https server up with external private key and certificates. 

This PR is introducing: 
* The setup functionality for a net/http webserver,
* The logic to load a user provided certificate to the webserver,
* Tests to validate the behaviour of all of these features,
* Command line options that will be useful in production.

The command line for GoCert accepts 1 mandatory argument:
* `--config` receives the path for a YAML file that describes the following variables:
    * KeyPath is the path for a private key of a x.509 certificate,
    * CertPath is the path for a x.509 certificate,
    * DBPath is the path for a file for SQLite to use,
    * Port is an integer for which port the service will be deployed in,

# Checklist:


Kayra1:
I've tried to be consistent with capitalization in errors and specifying RSA private key instead of just private key, but I may have slipped somewhere. Please let me know if you see any inconsistencies.

api/server.go
```diff
@@ -0,0 +1,59 @@
...
+// loadServerCertificates takes in a certificate and a private key and determines
+// whether to use self signed or given certificates, then returns it in a format
+// expected by the server
+func loadServerCertificates(certificate, key string) (*tls.Certificate, error) {
+	if certificate == "" || key == "" {
+		caCertPEM, caPrivateKeyPEM, err := certificates.GenerateCACertificate()
+		if err != nil {
+			return nil, err
+		}
+		certificate, key, err = certificates.GenerateSelfSignedCertificate(caCertPEM, caPrivateKeyPEM)
```
gruyaume, line 21: I'm not sure about generating self-signed certificates, I'd rather have the cert/key be mandatory parameters and error out if they are not provided. If this is requested in the future, we can add it then. But let's not add this preemptively.

gruyaume (changes requested):

api/server.go
```diff
@@ -0,0 +1,59 @@
...
+	"net/http"
+	"time"
+
+	"github.com/canonical/gocert/internal/certificates"
+)
+
+// loadServerCertificates takes in a certificate and a private key and determines
+// whether to use self signed or given certificates, then returns it in a format
+// expected by the server
+func loadServerCertificates(certificate, key string) (*tls.Certificate, error) {
```
ghislainbourgeois, line 15:
```suggestion
func loadServerCertificates(certificate, key string) (tls.Certificate, error) {
```
The documentation examples do not use pointers but values, I think we should follow that convention.

api/server.go
```diff
@@ -0,0 +1,59 @@
...
+func NewServer(certificate, key string) (*http.Server, error) {
+	serverCerts, err := loadServerCertificates(certificate, key)
+	if err != nil {
+		return nil, err
+	}
+	m := http.NewServeMux()
+	m.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
+		_, err := w.Write([]byte("Hello World"))
+		if err != nil {
+			return
```
ghislainbourgeois, line 43: Shouldn't this return nil and the error?

cmd/gocert/main.go
```diff
@@ -1,5 +1,36 @@
...
+	dbPathPtr := flag.String("db", "./certs.db", "The path of the SQLite database for the repository")
+	flag.Parse()
+
+	if *certPathPtr == "" || *keyPathPtr == "" {
+		fmt.Fprintf(os.Stderr, "Providing a certificate and matching private key is extremely highly recommended. To do so, run: --cert <path/to/certificate> --key <path/to/key>. Using self signed certificates.")
+	}
+	_, err := certdb.NewCertificateRequestsRepository(*dbPathPtr, "CertificateRequests")
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Couldn't connect to database: %s", err)
+		os.Exit(1)
```
ghislainbourgeois, line 24: For this construct, we could use the `log` module:
```suggestion
		logger.Fatalf(os.Stderr, "Couldn't connect to database: %s", err)
```

internal/certificates/certificates.go
```diff
@@ -0,0 +1,146 @@
...
+		return "", "", nil
+	}
+
+	certTemplate := &x509.Certificate{
+		SerialNumber: big.NewInt(1658),
+		Subject: pkix.Name{
+			Organization: []string{"Canonical, INC."},
+			Country:      []string{"US"},
+		},
+		IPAddresses:  []net.IP{net.IPv4(127, 0, 0, 1), net.IPv6loopback}, // TODO
```
ghislainbourgeois, line 80: I do not think we ever want to generate certificates for loopbacks.

internal/certificates/certificates.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+	return caCertPEM.String(), caPrivateKeyPEM.String(), nil
+}
+
+// GenerateSelfSignedCertificate will create a certificate and pk for GoCert coming from a rootCA
+// This certificate and PK is not saved anywhere.
+// This certificate and PK should either be saved somewhere, or a real certificate should be provided to GoCert
+func GenerateSelfSignedCertificate(caCertPEM, caPrivateKeyPEM string) (string, string, error) {
+	caCert, err := ParseCertificate(caCertPEM)
+	if err != nil {
```
ghislainbourgeois, line 66: Right now, the only usage we make of the CA certificate is to generate a self-signed certificate. I do not think we should convert the return values to string to parse them again in the next function.

In fact, I think we should mostly not convert to/from strings. When it becomes necessary, it should happen at the edges of the program.

internal/certificates/certificates.go
```diff
@@ -0,0 +1,146 @@
...
+		return "", "", nil
+	}
+	caPrivateKey, err := ParsePKCS1PrivateKey(caPrivateKeyPEM)
+	if err != nil {
+		return "", "", nil
+	}
+
+	certTemplate := &x509.Certificate{
+		SerialNumber: big.NewInt(1658),
```
ghislainbourgeois, line 75: This should not be hard-coded.

internal/certificates/certificates.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+	if block.Type != "CERTIFICATE" {
+		return *cert, errors.New("given PEM string not a certificate")
+	}
+	cert, err := x509.ParseCertificate(block.Bytes)
+	if err != nil {
+		return *cert, err
+	}
+	return *cert, nil
+}
```
ghislainbourgeois, line 129: Here, the documentation for `crypto/x509` uses pointers for `Certificate`.
```suggestion
func ParseCertificate(certPEM string) (*x509.Certificate, error) {
	block, _ := pem.Decode([]byte(certPEM))
	if block == nil {
		return nil, errors.New("PEM Certificate string not found or malformed")
	}
	if block.Type != "CERTIFICATE" {
		return nil, errors.New("given PEM string not a certificate")
	}
	cert, err := x509.ParseCertificate(block.Bytes)
	if err != nil {
		return nil, err
	}
	return cert, nil
}
```

internal/certificates/certificates.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+	if block.Type != "RSA PRIVATE KEY" {
+		return *pk, errors.New("given PEM string not an rsa private key")
+	}
+	pk, err := x509.ParsePKCS1PrivateKey(block.Bytes)
+	if err != nil {
+		return *pk, err
+	}
+	return *pk, nil
+}
```
ghislainbourgeois, line 146: Same thing here, we should use pointer semantics.
```suggestion
func ParsePKCS1PrivateKey(pkPEM string) (*rsa.PrivateKey, error) {
	block, _ := pem.Decode([]byte(pkPEM))
	if block == nil {
		return nil, errors.New("PEM private key string not found or malformed")
	}
	if block.Type != "RSA PRIVATE KEY" {
		return nil, errors.New("given PEM string not an rsa private key")
	}
	pk, err := x509.ParsePKCS1PrivateKey(block.Bytes)
	if err != nil {
		return nil, err
	}
	return pk, nil
}
```

ghislainbourgeois (changes requested):

cmd/gocert/main.go
```diff
@@ -1,5 +1,36 @@
...
+
+	server "github.com/canonical/gocert/api"
+	"github.com/canonical/gocert/internal/certdb"
+)
+
 func main() {
-	// ListenAndServe
+	certPathPtr := flag.String("cert", "", "A path for a certificate file to be used by the webserver")
+	keyPathPtr := flag.String("key", "", "The path for a private key for the given certificate")
+	dbPathPtr := flag.String("db", "./certs.db", "The path of the SQLite database for the repository")
```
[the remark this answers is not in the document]
Kayra1: Done

api/server.go
Kayra1: All gone

api/server.go
Kayra1: This function is gone

api/server.go
Kayra1: No, the http handler interface requires returning nothing.

cmd/gocert/main.go
Kayra1: Done. The logger doesn't support specifying the output writer in the `Fatalf` call itself, so I'm just setting the output to stderr in the beginning for now. This can have more advanced logic later.

internal/certificates/certificates.go
Kayra1: Gone

internal/certificates/certificates.go
Kayra1: This part is gone anyway but agreed

internal/certificates/certificates.go
Kayra1: Gone

internal/certificates/certificates.go
Kayra1: Gone

internal/certificates/certificates.go
Kayra1: also gone

api/server.go
```diff
@@ -0,0 +1,59 @@
...
+		}
+	})
+	s := &http.Server{
+		Addr: ":8080",
```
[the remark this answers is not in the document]
Kayra1: Done

cmd/gocert/main_test.go
```diff
@@ -0,0 +1,20 @@
...
+	// run gocert --cert openable path
+	// run gocert --key openable path
+	// run gocert --cert openable path --key openable path --db unopenable path
+	// run gocert --cert openable path --key openable path --db openable path
+	// parametrize below
+	// run gocert --cert unopenable path --key unopenable path
+	// run gocert --cert unopenable path --key openable path
+	// run gocert --cert openable path --key unopenable path
+	// run gocert --cert openable path --key openable path
+}
```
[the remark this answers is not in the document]
Kayra1: Not anymore

api/server.go
```diff
@@ -0,0 +1,90 @@
...
+	if err != nil {
+		return nil, err
+	}
+	router := http.NewServeMux()
+	router.HandleFunc("GET /", HelloWorld)
+
+	v1 := http.NewServeMux()
+	v1.Handle("/v1/", http.StripPrefix("/v1", router))
+
+	s := &http.Server{
```
gruyaume, line 70: Please avoid 1 letter variables

api/server.go
```diff
@@ -0,0 +1,90 @@
...
+		MaxHeaderBytes: 1 << 20,
+		TLSConfig: &tls.Config{
+			Certificates: []tls.Certificate{serverCerts},
+		},
+	}
+
+	return s, nil
+}
+
+func HelloWorld(w http.ResponseWriter, r *http.Request) {
```
gruyaume, line 85: Is this to run health checks?

Kayra1: No, it's just an example before we introduce real API calls. But that's also not a bad idea.

api/server.go
Kayra1: 1 letter variable names are actually [encouraged in some cases in Go](https://google.github.io/styleguide/go/decisions#single-letter-variable-names). `s` is actually the example variable name in Go's own docs [here](https://github.com/golang/go/blob/c71d2a8d8fcb50ed8ab0e1c4af211a0dcea16a75/src/net/http/doc.go#L76).

I can still change it if we don't prefer it.

gruyaume (approved):
looks good, make sure the pr description matches with its newer content. Here I'm referring to the fact we don't generate certs anymore.

api/server.go
ghislainbourgeois: Oh, my bad, I did not realize this was the return for the anonymous function.

api/server.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+	cert, err := os.ReadFile(c.CertPath)
+	if err != nil {
+		return config, errors.Join(validationErr, err)
+	}
+	key, err := os.ReadFile(c.KeyPath)
+	if err != nil {
+		return config, errors.Join(validationErr, err)
+	}
+	if _, err := os.OpenFile(c.DBPath, os.O_CREATE|os.O_RDONLY, 0644); err != nil {
```
ghislainbourgeois, line 48: If the file is opened without error, shouldn't we close it?

Kayra1: Good catch.

ghislainbourgeois (approved):

PR #7, opened 2024-04-16
feat: GoCert 1.0 Handlers

Kayra1:
# Description

This change introduces all of the handlers defined in the GoCert spec, and a helpful logging middleware function for the server.

# Checklist:


api/server.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	if _, err := os.OpenFile(c.DBPath, os.O_CREATE|os.O_RDONLY, 0644); err != nil {
+		return config, errors.Join(validationErr, err)
+	}
+	config.Cert = cert
+	config.Key = key
+	config.DBPath = c.DBPath
+	config.Port = c.Port
+	return config, nil
+}
```
gruyaume, line 63: Does this belong under `api`? It seems like it should be done before using api and only passing over what is needed.

gruyaume (changes requested):

api/server.go
```diff
@@ -0,0 +1,97 @@
...
+type Config struct {
+	Key    []byte
+	Cert   []byte
+	DBPath string
+	Port   int
+}
+
+type Environment struct {
+	DB *certdb.CertificateRequestsRepository
+}
```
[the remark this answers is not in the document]
Kayra1: I actually had started off with db directly, but when I searched online for different ways to inject variables into handlers, people were doing the pattern that I was doing for environment variables since there's no global environment variable. Right now it's just DB, but it could be any environment variable that the handlers may need to be dynamically loaded, for example some config values from the configuration file like constraints to the certs could be placed here.

cmd/gocert/main.go
```diff
@@ -1,5 +1,27 @@
...
+	"flag"
+	"log"
+	"os"
+
+	server "github.com/canonical/gocert/api"
+)
+
 func main() {
-	// ListenAndServe
+	log.SetOutput(os.Stderr)
```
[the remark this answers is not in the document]
Kayra1: Yes

api/handlers.go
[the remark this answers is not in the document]
Kayra1: Because I think the api itself will be consumed by the charm, an external program, and there's precedent of doing that in the second example from the [project-layout](https://github.com/golang-standards/project-layout/tree/master/api) repo, [moby](https://github.com/moby/moby/tree/master/api). 

I can change it if we prefer it that way.

api/server.go
Kayra1: I think there could be a better place for this, but as it stands the config file is for configuring the server, and because it's very difficult to share variables between tests, it's more convenient to keep them together. If we want to move both the server and the config for the server somewhere else, I'm open to that.

cmd/gocert/main.go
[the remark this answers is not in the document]
gruyaume: Why not log to stdout?

api/handlers.go
```diff
@@ -0,0 +1,156 @@
...
+	"io"
+	"log"
+	"net/http"
+	"strconv"
+	"strings"
+)
+
+// GoCertRouter takes in an environment struct, passes it along to any handlers that will need
+// access to it, then builds and returns it for a server to consume
+func GoCertRouter(env *Environment) http.Handler {
```
ghislainbourgeois, line 15: Small nitpick, should this be named `NewGoCertRouter`?

api/handlers.go
```diff
@@ -0,0 +1,156 @@
...
+	v1 := http.NewServeMux()
+	v1.HandleFunc("GET /status", HealthCheck)
+	v1.Handle("/api/v1/", http.StripPrefix("/api/v1", router))
+
+	return logging(v1)
+}
+
+// the health check endpoint simply returns a 200
+func HealthCheck(w http.ResponseWriter, r *http.Request) {
+	w.WriteHeader(200) //nolint:errcheck
```
ghislainbourgeois, line 32:
```suggestion
	w.WriteHeader(http.StatusOK) //nolint:errcheck
```

I would suggest the same kind of change everywhere we use status code, including the test files.

api/handlers.go
```diff
@@ -0,0 +1,156 @@
...
+
+// GetCertificateRequests receives an id as a path parameter, and
+// returns the corresponding Certificate Request
+func GetCertificateRequest(env *Environment) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		id := r.PathValue("id")
+		cert, err := env.DB.Retrieve(id)
+		if err != nil {
+			if err.Error() == "sql: no rows in result set" {
+				logError(err.Error(), 400, w)
```
ghislainbourgeois, line 87: Shoulnd't we return 404 here?

api/handlers.go
```diff
@@ -0,0 +1,156 @@
...
+	return func(w http.ResponseWriter, r *http.Request) {
+		cert, err := io.ReadAll(r.Body)
+		if err != nil {
+			logError(err.Error(), 400, w)
+			return
+		}
+		id := r.PathValue("id")
+		insertId, err := env.DB.Update(id, string(cert))
+		if err != nil {
+			logError(err.Error(), 400, w)
```
ghislainbourgeois, line 130: Should we return 404 here instead?

api/handlers.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+// DeleteCertificateRequest handler receives an id as a path parameter,
+// deletes the corresponding Certificate Request, and returns a 204 on success
+func DeleteCertificateRequest(env *Environment) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		id := r.PathValue("id")
+		err := env.DB.Delete(id)
+		if err != nil {
+			logError(err.Error(), 500, w)
```
ghislainbourgeois, line 111: Do you know if an error is returned by the DB if we try to delete something that does not exist? In that case, we should not return a 500, but something either a success, or something in the 4xx range.

api/handlers.go
```diff
@@ -0,0 +1,156 @@
...
+			return
+		}
+		id, err := env.DB.Create(string(csr))
+		if err != nil {
+			if strings.Contains(err.Error(), "UNIQUE constraint failed") {
+				logError("given csr already recorded", 400, w)
+				return
+			} else {
+				logError(err.Error(), 400, w)
```
ghislainbourgeois, line 68: I think this might be good enough for now, but I assume some errors would be due to the content of the request, and some would be due to something on our side, like the DB not responding for some reason. We might have to revisit this to return a 5xx if the error is on our side. Additionally, if the error is with the content sent to us, we could return 415 or 422.

This could be out of scope for this PR however and evaluated as part of another task.

ghislainbourgeois (changes requested):

api/handlers.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+		w.WriteHeader(204)
+	}
+}
+
+// PostCertificate handler receives an id as a path parameter,
+// and attempts to add a given certificate to the corresponding certificate request
+func PostCertificate(env *Environment) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		cert, err := io.ReadAll(r.Body)
```
gruyaume, line 122: Should we validate the content of the body? Here it looks like anything can be POSTed to the endpoint.

api/handlers.go
```diff
@@ -0,0 +1,156 @@
...
+
+// logError is a helper function that logs any error and writes it back as an http response
+func logError(msg string, status int, w http.ResponseWriter) {
+	errMsg := fmt.Sprintf("error: %s", msg)
+	log.Println(errMsg)
+	w.WriteHeader(status)
+	if _, err := w.Write([]byte(errMsg)); err != nil {
+		logError(err.Error(), 500, w)
+	}
+}
```
gruyaume, line 156: The name does not match with the content. Logging an error vs returning an error status code in the http request is not the same thing.

api/handlers.go
```diff
@@ -0,0 +1,156 @@
...
+	router.HandleFunc("POST /certificate_requests", PostCertificateRequest(env))
+	router.HandleFunc("GET /certificate_requests/{id}", GetCertificateRequest(env))
+	router.HandleFunc("DELETE /certificate_requests/{id}", DeleteCertificateRequest(env))
+	router.HandleFunc("POST /certificate_requests/{id}/certificate", PostCertificate(env))
+
+	v1 := http.NewServeMux()
+	v1.HandleFunc("GET /status", HealthCheck)
+	v1.Handle("/api/v1/", http.StripPrefix("/api/v1", router))
+
+	return logging(v1)
```
gruyaume, line 27: How will this scale when we add more middlewares (Ex. for authentication)? Does returning logging still make sense?

api/handlers.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+// GetCertificateRequests receives an id as a path parameter, and
+// returns the corresponding Certificate Request
+func GetCertificateRequest(env *Environment) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		id := r.PathValue("id")
+		cert, err := env.DB.Retrieve(id)
+		if err != nil {
+			if err.Error() == "sql: no rows in result set" {
```
gruyaume, line 86: We shouldn't have anything sql-y here, everything db related should be in `certdb`. I.e. we should be able to switch to nosql without changing this file at all.

gruyaume (changes requested):

api/handlers.go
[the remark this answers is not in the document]
Kayra1: The API and not the module is exposed in moby also, but I've changed it since we prefer it there.

api/handlers.go
Kayra1: makes sense

api/handlers.go
Kayra1: Done

api/handlers.go
Kayra1: I think 404 could be confusing, since it means that the path isn't available. Users may assume that the endpoint is wrong rather than the request parameters. Maybe a 204 since the request was valid?

api/handlers.go
Kayra1: I'd prefer not to have explicit 404's in general, go will return a 404 for paths that don't exist. In this case, if there's an error, it will be either because of the validation of the cert or a connection error. I've differentiated between the errors in the certdb side so this part reads a bit cleaner.

api/handlers.go
Kayra1: No, if nothing is deleted there is no error. It's still a good idea to differentiate this, so I've added a custom error.

api/handlers.go
Kayra1: This line is a real error, thanks for catching it.

I've done my best to differentiate the two cases, but excluding the errors I couldn't forsee, I think these responses are "production" "ready". In my experience it's generally a bad idea to use rarely used 400's because the consumers of API's like to switch over error messages. When it comes to errors about the HTTP request directly, Go will send the proper expected responses in those situations before it ever gets to the handler code.

Like you said though, we can always come back to this later.

Maybe we can send a 418 for the health check :D

api/handlers.go
Kayra1: `db.Update` does the validation for both the cert and the id given in the URL. We could do extra things here, but the common "PEM Certificate string not found or malformed" response is enough imo.

I added a test that shows this response.

api/handlers.go
Kayra1: agreed. changed the name to be more precise.

api/handlers.go
Kayra1: we can do something like

`
stack := createMiddlewareStack(
         Logging,
         Authentication,
         Metrics,
         ValidatePermissions,        
)
return stack(v1)
`

api/handlers.go
Kayra1: cleaned up the errors in many places, and wrapped this error in a new custom string in `certdb`.

Kayra1:
The first PR got merged, so I rebased this branch. Should be easier to read now.

internal/api/handlers.go
```diff
@@ -0,0 +1,172 @@
...
+
+// NewGoCertRouter takes in an environment struct, passes it along to any handlers that will need
+// access to it, then builds and returns it for a server to consume
+func NewGoCertRouter(env *Environment) http.Handler {
+	router := http.NewServeMux()
+	router.HandleFunc("GET /certificate_requests", GetCertificateRequests(env))
+	router.HandleFunc("POST /certificate_requests", PostCertificateRequest(env))
+	router.HandleFunc("GET /certificate_requests/{id}", GetCertificateRequest(env))
+	router.HandleFunc("DELETE /certificate_requests/{id}", DeleteCertificateRequest(env))
+	router.HandleFunc("POST /certificate_requests/{id}/certificate", PostCertificate(env))
```
gruyaume, line 21: It looks like we are missing the corresponding DELETE to `/certificate_requests/{req-id}/certificate/{cert-id}`

Kayra1: hmm, I copied over the methods from the spec. I will add this to the spec and then make the corresponding request available.

api/handlers.go
ghislainbourgeois: Please no!

internal/api/handlers.go
Kayra1: added the feature to certdb and handlers, and added tests for everything new.

ghislainbourgeois (approved):

gruyaume (approved):

PR #11, opened 2024-04-26
feat: add nextjs skeleton for the ui

Kayra1:
# Description

## Main UI Code

This is a nextjs UI skeleton which is served by the gocert application located in the UI folder. Simply do:
`bun i` / `npm i`
and
`bun run build` / `npm run build`

to produce an `out/` folder that contains the static assets for the frontend. This directory is then served up.

## CI for the frontend

On top of the usual CI, there is now a step to build the website to make sure that the rock will build without issue, and there is another step to run the component tests for the nextjs application.

### Build

Building uses the mechanism nextjs provides. There is first class IDE support for this option.

### Testing

The nextjs app is tested using vitest, which does component level tests. Each test output reports directly to github actions, giving embedded test response messages:

image

## The ROCK

The ROCK now has an extra step to build the UI from within gocert in the build step, and go's installer will copy over the necessary assets.

# Checklist:


internal/api/handlers.go
gruyaume, line 1: We are missing the changes in the rock

ui/README.md
```diff
@@ -0,0 +1,36 @@
...
+- [Next.js Documentation](https://nextjs.org/docs) - learn about Next.js features and API.
+- [Learn Next.js](https://nextjs.org/learn) - an interactive Next.js tutorial.
+
+You can check out [the Next.js GitHub repository](https://github.com/vercel/next.js/) - your feedback and contributions are welcome!
+
+## Deploy on Vercel
+
+The easiest way to deploy your Next.js app is to use the [Vercel Platform](https://vercel.com/new?utm_medium=default-template&filter=next.js&utm_source=create-next-app&utm_campaign=create-next-app-readme) from the creators of Next.js.
+
+Check out our [Next.js deployment documentation](https://nextjs.org/docs/deployment) for more details.
```
gruyaume, line 36:
```suggestion
```

ui/README.md
```diff
@@ -0,0 +1,36 @@
...
+yarn dev
+# or
+pnpm dev
+# or
+bun dev
+```
+
+Open [http://localhost:3000](http://localhost:3000) with your browser to see the result.
+
+You can start editing the page by modifying `app/page.tsx`. The page auto-updates as you edit the file.
```
gruyaume, line 19:
```suggestion
You can start editing the page by modifying `src/app/page.tsx`. The page auto-updates as you edit the file.
```

ui/public/next.svg
gruyaume, line 1: Do we want this image?

gruyaume (changes requested):

internal/api/handlers.go
Kayra1: This should have been a draft PR, but I can't seem to switch it back. I'm working on it right now.

ui/public/next.svg
Kayra1: No

internal/api/handlers.go
gruyaume: Ok I'll convert it to a draft

Kayra1: ty

ui/README.md
Kayra1: gone

ui/public/next.svg
Kayra1: its gone

internal/api/server.go
```diff
@@ -15,21 +15,24 @@ import (
...
 
 type ConfigYAML struct {
-	KeyPath  string
-	CertPath string
-	DBPath   string
-	Port     int
+	KeyPath     string
+	CertPath    string
+	DBPath      string
+	FrontendDir string
```
Kayra1, line 21: I've made it possible to provide any directory for the frontend, but I'm not sure if this is a good idea.

ghislainbourgeois (approved):

.github/workflows/main.yml
saltiyazan, line 1: This isn't the scope of this PR, but since we are at adding new jobs, how about extracting them to separate files like we do in other projects?

ui/README.md
saltiyazan, line 1: Should we add a reference in the main README file to this?

.github/workflows/main.yml
```diff
@@ -20,7 +20,22 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: Use Bun
+        uses: oven-sh/setup-bun@v1
+        with:
+          bun-version: latest
 
+      - name: Build frontend
+        run: |
+          bun install
```
gruyaume, line 37: What si this `bun` thing? Do we really need it? Why add this dependency in addition to next?

rockcraft.yaml
gruyaume, line 1: Have you tested the rock and it works fine?

Kayra1: Yes, I followed [this](https://canonical-rockcraft.readthedocs-hosted.com/en/latest/tutorials/hello-world/) hello world rock tutorial but packed and deployed this rock, and it worked.

.github/workflows/main.yml
Kayra1: It is a faster alternative to npm. We can use that too, but npm's action was complaining about some random dependency issue that I couldn't resolve. npm, bun, deno and a bunch of others are all javascript runtimes but all use the same v8 engine in chrome for javascript, so they produce the same javascript code but their tooling around transpiling is a bit different.

Kayra1: It is not a dependency in gocert, neither the backend nor the frontend, it is only in the workflow, and installs the package.json and runs next build just like node would.

ui/README.md
Kayra1: A reference to what?

.github/workflows/main.yml
Kayra1: I am not familiar with how to do that, I can do it in the next PR.

.github/workflows/main.yml
gruyaume: I'm not confident about adding this dependency to fix another issue. Let's try to fix the issue. Plus we're not even using `bun` when running building or running the service in the rock.

.github/workflows/main.yml
gruyaume, line 1: I'd recommend adding an integration tests that takes the built rock, runs it, and some code that checks that the UI comes up.

Similar to what we do here:
- https://github.com/canonical/sdcore-nms/blob/main/.github/workflows/integration_tests.yaml

gruyaume (changes requested):

.github/workflows/main.yml
Kayra1: I will work on this

.github/workflows/main.yml
Kayra1: gone

rockcraft.yaml
Kayra1: Added a bit more detail to contributing.md too

.github/workflows/main.yml
saltiyazan: Sure :)

ui/README.md
Kayra1: makes sense

Kayra1: added it to contributing.md

.github/workflows/main.yml
Kayra1: I ended up having to modify the actions a lot so i've split them into files here

Kayra1: I added this integration test to the build-rock step, and made some changes to the frontend loading code to embed the frontend files into the executable so we don't have to carry the ui/out directory alongside it everywhere.

FYI, the action in the link you've sent has never ran before, and it probably wouldn't work because curl outputs the response body to stderr, it would probably fail to find 5G NMS in the output and always fail.

CONTRIBUTING.md
```diff
@@ -2,24 +2,37 @@
...
 ## Getting Started
 
+1. Install Go and Nodejs
 1. Fork the repository on GitHub
 2. Clone the forked repository to your local machine
-3. Build the project: `go build ./...`
-4. Run the project: `./gocert`
+3. Build the frontend: `cd ui && npm build build`
+3. Install the project: `go install ./...`
+4. Run the project: `gocert --config config.yaml`
```
gruyaume, line 10: This `config.yaml` does not exist? Should we include an example one in the project?

gruyaume (changes requested):

ui/bun.lockb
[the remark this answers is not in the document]
Kayra1: gone

ui/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "gocert",
+  "version": "0.1.0",
```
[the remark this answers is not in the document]
Kayra1: Should it? I interpreted this as the version of the frontend, and the rockcraft.yaml version as the version of the OCI image.

CONTRIBUTING.md
Kayra1: No, we could either add another step to create a cert.pem, key.pem and the config.yaml, or include example files of each if we want this to work out of the box.

gruyaume: I believe ncluding an example `config.yaml` would be useful. The `.pem` files are not as useful though, those are quite standard files.

ui/package.json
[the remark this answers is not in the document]
gruyaume: I think we can sort of ignore the frondend version then. And only treat the go version as the `gocert` version. The two (frontend and backend) will be packaged together.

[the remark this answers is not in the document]
Kayra1: that's fine with me. one less thing to track.

CONTRIBUTING.md
Kayra1: what about just putting the contents of the config.yaml in this file? I don't want to add files to the top level folder if i can avoid it.

Kayra1: this is how it looks: https://github.com/canonical/gocert/blob/TLSENG-226-frontend-skeleton/CONTRIBUTING.md

gruyaume: If we want to embed it in another file, I'd suggest using the README.md as it will not only be useful to contributors but to any other users as well.

Kayra1: moved it there

gruyaume (approved):

PR #12, opened 2024-04-30
feat: exposes baseline metrics

saltiyazan:
# Description

- Exposes basic Go metrics on the Metric endpoint
- Adds a test case
- Improves configYaml struct to specify required keys

The solution creates a custom handler for the `metrics` endpoint, that handler will run for any collectors registered in the metrics registry, for now we only have the collector of some basic Go metrics, but this should be easy to extend in the future with custom collectors.

To test this, run the service and curl the `/metrics` endpoint.
To test this with prometheus, configure prometheus with the following job:
```
scrape_configs:
  - job_name: gocert
    scrape_interval: 10s
    static_configs:
    - targets:
      - localhost:443
```

Run prometheus then curl the metrics endpoint of prometheus, for example:
`curl http://localhost:9090/metrics | grep gocerts'

# Checklist:


internal/api/handlers.go
```diff
@@ -25,6 +29,10 @@ func NewGoCertRouter(env *Environment) http.Handler {
 	v1 := http.NewServeMux()
 	v1.HandleFunc("GET /status", HealthCheck)
 	v1.Handle("/api/v1/", http.StripPrefix("/api/v1", router))
+	reg := prometheus.NewRegistry()
+	reg.MustRegister(collectors.NewGoCollector())
+	prometheusHandler := promhttp.HandlerFor(reg, promhttp.HandlerOpts{})
+	v1.Handle("/api/v1/metrics", prometheusHandler)
```
gruyaume, line 35: I'd recommend treating metrics as their own independent web server and not placing them under `internal/api/handlers.go`. Instead, they can be under `/metrics/metrics.go` and directly started from `main` via something like `go metrics.Init()`.

## Rationale
1. The metrics server is independent from the other gocert specific endpoints and can be maintain on its own. It does not make sense for it to be under `api/v1` in my view, it's not part of the gocert api.

2. If for whatever reason the technology we use for metrics changes from Prometheus to something else, all we need to change is `internal/metrics`, just like we do for the db.

gruyaume (changes requested):

internal/api/handlers.go
Kayra1: While it may make sense to have a separate server for collecting metrics for the go application, it will severely limit the information we can get from the GoCert server. Unless we manually watch stdout, stderr and the db, we can not collect any metrics regarding the HTTP requests or responses, or what action the server has taken, only what it can infer from external data.

I do agree that it should not be under `api/v1`, and agree that this implementation is too coupled to prometheus.

I recommend writing a `Metrics` function that returns an object that implements the http.Handler interface, which can be swapped out for any other metric provider while keeping the same level of access to the GoCert API server.

internal/api/handlers.go
```diff
@@ -25,6 +29,10 @@ func NewGoCertRouter(env *Environment) http.Handler {
 	v1 := http.NewServeMux()
 	v1.HandleFunc("GET /status", HealthCheck)
 	v1.Handle("/api/v1/", http.StripPrefix("/api/v1", router))
+	reg := prometheus.NewRegistry()
+	reg.MustRegister(collectors.NewGoCollector())
+	prometheusHandler := promhttp.HandlerFor(reg, promhttp.HandlerOpts{})
```
Kayra1, line 34: Code relevant to prometheus should be abstracted to its own internal function. It would be nicer to have a function simply return `http.Handler`, which promhttp.HandlerFor already returns, so we can keep the metrics uncoupled.

Kayra1 (changes requested):

saltiyazan:
Converted to draft, figuring out why are unit tests failing

internal/api/handlers.go
saltiyazan: I created a metrics package, there an abstract metrics handler and a prometheus metrics handler.
I renamed the Mux from v1.

Let me know what you think. @gruyaume @Kayra1

internal/api/handlers.go
```diff
@@ -22,11 +23,15 @@ func NewGoCertRouter(env *Environment) http.Handler {
...
-	v1.HandleFunc("GET /status", HealthCheck)
-	v1.Handle("/api/v1/", http.StripPrefix("/api/v1", router))
+	monitoringMux := http.NewServeMux()
+	monitoringMux.HandleFunc("/status", HealthCheck)
+	monitoringMux.Handle("/metrics", metricsHandler)
 
-	return logging(v1)
+	api := http.NewServeMux()
+	api.Handle("/api/v1/", http.StripPrefix("/api/v1", router))
+	api.Handle("/", monitoringMux)
```
gruyaume, line 32: Do we need a new mux for minitoring, here we could simply call 

```
router.Handle("/metrics", NewMetricsHandler())
```

internal/api/server.go
```diff
@@ -85,7 +86,8 @@ func NewServer(configFile string) (*http.Server, error) {
 
 	env := &Environment{}
 	env.DB = db
-	router := NewGoCertRouter(env)
+	metricsHandler := metrics.NewPrometheusHandler()
```
gruyaume, line 89: I don't think we need to make any change in this file

internal/metrics/metrics.go
```diff
@@ -0,0 +1,32 @@
...
+	registry.MustRegister(collectors.NewGoCollector())
+	return &PrometheusHandler{
+		registry: registry,
+	}
+}
+
+// ServeHTTP implements the http.Handler interface, allowing the PrometheusHandler to handle HTTP requests.
+func (p *PrometheusHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	promhttp.HandlerFor(p.registry, promhttp.HandlerOpts{}).ServeHTTP(w, r)
+}
```
gruyaume, line 32: Do we need all of this? With my recommendation above, we could have something like

```
func NewMetricsHandler() http.Handler {
	reg := prometheus.NewRegistry()
	reg.MustRegister(collectors.NewGoCollector())
	prometheusHandler := promhttp.HandlerFor(reg, promhttp.HandlerOpts{})
	return prometheusHandler
}
```

saltiyazan: What you are proposing would definitely be sufficient right now, however the aim was to have a way of implementing metrics handlers other than the one for Prometheus.
What do you think?

internal/api/handlers.go

internal/api/server.go
saltiyazan: Done

internal/metrics/metrics.go
saltiyazan: Done

gruyaume (approved):
Looks good, the description should be adapted to the new approach (i.e. metrics are not under `/api/v1`.

I tested this locally and it looks good :)

image

internal/api/handlers_test.go
```diff
@@ -336,8 +344,15 @@ func TestGoCertRouter(t *testing.T) {
...
-			if res.StatusCode != tC.status || string(resBody) != tC.response {
-				t.Errorf("expected response did not match.\nExpected vs Received status code: %d vs %d\nExpected vs Received body: \n%s\nvs\n%s\n", tC.status, res.StatusCode, tC.response, string(resBody))
+			switch path := tC.path; path {
+			case "/metrics":
+				if res.StatusCode != tC.status || !strings.Contains(string(resBody), tC.response) {
+					t.Errorf("expected response did not match.\nExpected vs Received status code: %d vs %d\nExpected vs Received body: \n%s\nvs\n%s\n", tC.status, res.StatusCode, tC.response, string(resBody))
+				}
+			default:
+				if res.StatusCode != tC.status || string(resBody) != tC.response {
+					t.Errorf("expected response did not match.\nExpected vs Received status code: %d vs %d\nExpected vs Received body: \n%s\nvs\n%s\n", tC.status, res.StatusCode, tC.response, string(resBody))
```
Kayra1, line 354: You can remove the switch statement and just use `strings.Contains` for either case, since it will return true if the strings match precisely too.

Kayra1 (changes requested):
perfect. just one small thing.

internal/api/handlers_test.go
saltiyazan: Done, thanks

Kayra1 (approved):

PR #16, opened 2024-05-13
feat: Expose certificate metrics

kayra1:
# Description

This change introduces the collection of metrics for GoCert. These metrics are the ones described in the GoCert backend specification, and two metrics specific to the webserver: Total number of requests and Request Latency.

There are 2 main methods of collecting these metrics: 
* As middleware that can sit between the endpoint handlers and the request and responses, and record anything that happens.
* As a separate goroutine that periodically queries the DB to update the metrics.

There are pros and cons to doing either.

### Middleware

For the middleware method, the main pro is that there is visibility to every process the webserver does, whether that's auth, logs, or anything in the request and response objects. This is nice for collecting information about the webserver, like number of failed logins, IP addresses of the requests, internal server errors etc.

It is not a good idea to collect metrics about things where the source of truth is not the requests themselves. This couples the API to the metrics themselves (how do we know the deleted CSR was outstanding or not without passing this info specifically?), introduces complex code to correctly identify the resource being edited and may lead to desync issues with the data in the database and the metrics (if we fail to pass the outstanding or not information to the middleware).

### GoRoutine

For the goroutine method, the benefit is the fact that the metrics are collected from the source of truth, which means the metrics are always eventually consistent. It's also completely decoupled from the metrics endpoint, which means we get to make decisions about moving the handler to different ports or servers without messing with the metrics collection logic and vice versa.

The downside is, the cadence of the collection means there will need to be a balance between how up-to-date the metrics are at any given time and the amount of load and processing power we need to use from the host and the DB to collect these metrics. Considering the fact that certificates are low in number and the metrics for them change very infrequently, this is not a problem even if we choose to collect metrics every 6 hours.

This PR initializes the cadance value at 120 seconds. This can be set as a config option in the future.

In this PR there's examples of doing both, which should be chosen considering these factors when defining future metrics.

# Checklist:

### DIFF
--- a/internal/api/middleware.go
+++ b/internal/api/middleware.go
@@ -3,13 +3,17 @@ package server
 import (
 	"log"
 	"net/http"
+	"strings"
+
+	"github.com/canonical/gocert/internal/metrics"
 )
 
 type Middleware func(http.Handler) http.Handler
 
-// The Context type helps middleware pass along information through the chain.
-type Context struct {
+// The MiddlewareContext type helps middleware pass along information through the chain.
+type MiddlewareContext struct {
 	responseStatusCode int
+	metrics            *metrics.PrometheusMetrics
 }
 
 // The ResponseWriterCloner struct implements the http.ResponseWriter class, and copies the status
@@ -30,20 +34,41 @@ func (rwc *ResponseWriterCloner) WriteHeader(code int) {
 	rwc.ResponseWriter.WriteHeader(code)
 }
 
+// createMiddlewareStack chains given middleware for the server.
+// Each middleware functions calls next.ServeHTTP in order to resume the chain of execution.
+// The order these functions are given to createMiddlewareStack matters.
+// The functions will run the code before next.ServeHTTP in order.
+// The functions will run the code after next.ServeHTTP in reverse order.
+func createMiddlewareStack(middleware ...Middleware) Middleware {
+	return func(next http.Handler) http.Handler {
+		for i := len(middleware) - 1; i >= 0; i-- {
+			mw := middleware[i]
+			next = mw(next)
+		}
+		return next
+	}
+}
+
 // The Metrics middleware captures any request relevant to a metric and records it for prometheus.
-func Metrics(ctx *Context) Middleware {
+func Metrics(ctx *MiddlewareContext) Middleware {
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			next.ServeHTTP(w, r)
-			if ctx.responseStatusCode != 200 {
+			if ctx.responseStatusCode/100 != 2 {
 				return
+			}
+			if r.Method == "POST" && r.URL.Path == "/api/v1/certificate_requests" {
+				ctx.metrics.CertificateRequests.Inc()
+			}
+			if r.Method == "DELETE" && strings.HasPrefix(r.URL.Path, "/api/v1/certificate_requests") {
+				ctx.metrics.CertificateRequests.Dec()
 			}
 		})
 	}
 }
 
 // The logging middleware captures any http request coming through, and logs it.
-func Logging(ctx *Context) Middleware {
+func Logging(ctx *MiddlewareContext) Middleware {
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			clonedWwriter := NewResponseWriter(w)
--- a/internal/api/middleware.go
+++ b/internal/api/middleware.go
@@ -3,9 +3,9 @@ package server
 import (
 	"log"
 	"net/http"
-	"strings"
 
 	"github.com/canonical/gocert/internal/metrics"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
 type Middleware func(http.Handler) http.Handler
@@ -55,16 +55,14 @@ func createMiddlewareStack(middleware ...Middleware) Middleware {
 func Metrics(ctx *MiddlewareContext) Middleware {
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-			next.ServeHTTP(w, r)
-			if ctx.responseStatusCode/100 != 2 {
-				return
-			}
-			if r.Method == "POST" && strings.HasSuffix(r.URL.Path, "/certificate_requests") {
-				ctx.metrics.CertificateRequests.Inc()
-			}
-			if r.Method == "DELETE" && strings.HasSuffix(r.URL.Path, "/certificate_requests") {
-				ctx.metrics.CertificateRequests.Dec()
-			}
+			base := promhttp.InstrumentHandlerCounter(
+				&ctx.metrics.RequestsTotal,
+				promhttp.InstrumentHandlerDuration(
+					&ctx.metrics.RequestsDuration,
+					next,
+				),
+			)
+			base.ServeHTTP(w, r)
 		})
 	}
 }
--- a/internal/api/middleware.go
+++ b/internal/api/middleware.go
@@ -1,0 +1,55 @@
+package server
+
+import (
+	"log"
+	"net/http"
+)
+
+type Middleware func(http.Handler) http.Handler
+
+// The Context type helps middleware pass along information through the chain.
+type Context struct {
+	responseStatusCode int
+}
+
+// The ResponseWriterCloner struct implements the http.ResponseWriter class, and copies the status
+// code of the response for the middleware to be able to read the responses.
+type ResponseWriterCloner struct {
+	http.ResponseWriter
+	statusCode int
+}
+
+// NewResponseWriter returns a new ResponseWriterCloner struct
+func NewResponseWriter(w http.ResponseWriter) *ResponseWriterCloner {
+	return &ResponseWriterCloner{w, http.StatusOK}
+}
+
+// WriteHeader duplicates the status code into the cloner struct for reading
+func (rwc *ResponseWriterCloner) WriteHeader(code int) {
+	rwc.statusCode = code
+	rwc.ResponseWriter.WriteHeader(code)
+}
+
+// The Metrics middleware captures any request relevant to a metric and records it for prometheus.
+func Metrics(ctx *Context) Middleware {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			next.ServeHTTP(w, r)
+			if ctx.responseStatusCode != 200 {
+				return
+			}
+		})
+	}
+}
+
+// The logging middleware captures any http request coming through, and logs it.
+func Logging(ctx *Context) Middleware {
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			clonedWwriter := NewResponseWriter(w)
+			next.ServeHTTP(w, r)
+			log.Println(r.Method, r.URL.Path, clonedWwriter.statusCode, http.StatusText(clonedWwriter.statusCode))
+			ctx.responseStatusCode = clonedWwriter.statusCode
+		})
+	}
+}
--- a/internal/api/middleware.go
+++ b/internal/api/middleware.go
@@ -10,35 +10,37 @@ import (
 
 type Middleware func(http.Handler) http.Handler
 
-// The MiddlewareContext type helps middleware pass along information through the chain.
+// The MiddlewareContext type helps middleware receive and pass along information through the middleware chain.
 type MiddlewareContext struct {
 	responseStatusCode int
 	metrics            *metrics.PrometheusMetrics
 }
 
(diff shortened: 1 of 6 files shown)


internal/metrics/metrics.go
```diff
@@ -1,17 +1,224 @@
...
+	}
+	metricsBackend := newPrometheusMetrics()
+	metricsBackend.generateMetrics(csrs)
+	go func() {
+		time.Sleep(120 * ONE_SECOND)
+		metricsBackend.generateMetrics(csrs)
+	}()
+	metricsBackend.Handler = promhttp.HandlerFor(metricsBackend.registry, promhttp.HandlerOpts{})
+	return metricsBackend
+}
```
gruyaume, line 49: Should we use a ticker instead of continuously running the goroutine?

Ex:

```go
ticker := time.NewTicker(120 * time.Second)
go func() {
    for range ticker.C {
        metricsBackend.generateMetrics(csrs)
    }
}()
```

This will also remove the need of defining and using the `ONE_SECOND` const.

internal/metrics/metrics.go
```diff
@@ -1,17 +1,224 @@
...
+			Buckets: prometheus.ExponentialBuckets(0.1, 1.5, 5),
+		}, []string{"method", "code"},
+	)
+	return *metric
+}
+
+func certificateExpiryDate(certString string) time.Time {
+	certBlock, _ := pem.Decode([]byte(certString))
+	cert, _ := x509.ParseCertificate(certBlock.Bytes)
+	// TODO: cert.NotAfter can exist in a wrong cert. We should catch that at the db level validation
```
gruyaume, line 222: There are a bunch of TODO's in this PR, how are we tracking those?

internal/metrics/metrics.go
```diff
@@ -1,17 +1,224 @@
...
+			}
+			if daysRemaining < 30 {
+				expiringIn30DaysCertCount += 1
+			}
+			if daysRemaining < 90 {
+				expiringIn90DaysCertCount += 1
+			}
+		}
+	}
+	pm.CertificateRequests.Set(float64(csrCount))
```
gruyaume, line 125: Why do we do those type conversions instead of setting `csrCount` to be a float64 from the start?

internal/api/middleware.go
```diff
@@ -0,0 +1,80 @@
...
+	responseStatusCode int
+	metrics            *metrics.PrometheusMetrics
+}
+
+// The responseWriterCloner struct wraps the http.ResponseWriter struct, and extracts the status
+// code of the response writer for the middleware to read
+type responseWriterCloner struct {
+	http.ResponseWriter
+	statusCode int
+}
```
gruyaume, line 24: I'm somewhat confused by this approach. Is this common practice?

gruyaume (commented):
This is a quick preliminary review, I don't have a strong opinion on the general approach to metrics (middleware vs separate goroutine). Is there a common approach to this problem? What do major go projects do?

Kayra1:
> Is there a common approach to this problem? What do major go projects do?

The prom_client library has both options as examples for [goroutines](https://github.com/prometheus/client_golang/blob/36b9f46811b73b12b09ca9b6cf65bc04291db555/examples/exemplars/main.go#L48) and [middleware](https://github.com/prometheus/client_golang/blob/36b9f46811b73b12b09ca9b6cf65bc04291db555/examples/middleware/httpmiddleware/httpmiddleware.go#L74). We are using both approaches for the metrics that are relevant for each method.

internal/metrics/metrics.go
Kayra1: Tickers look nice, I've changed it to use this.

internal/metrics/metrics.go
Kayra1: I can make a JIRA ticket for them for the future, I don't think it's relevant to this PR but it should not be ignored.

internal/metrics/metrics.go
Kayra1: I didn't even realize VS code automatically put type conversions here. Changed.

internal/api/middleware.go
Kayra1: Wrapping functions and implementing stdlib interfaces I've found to be very common in Go programs. There are some tutorials that also show this as an example for being able to read from writers. There are also a bunch of native functions for this like io.MultiWriter and io.TeeReader, but they required modifying the handlers themselves which I wanted to avoid for coupling reasons.

internal/metrics/metrics.go
gruyaume: Ok, please do

Kayra1: TLSENG-243 created

internal/metrics/metrics.go
```diff
@@ -1,17 +1,224 @@
...
+	CertificatesExpiringIn90Days   prometheus.Gauge
+	ExpiredCertificates            prometheus.Gauge
+
+	RequestsTotal    prometheus.CounterVec
+	RequestsDuration prometheus.HistogramVec
+}
+
+// NewMetricsSubsystem returns the metrics endpoint HTTP handler and the Prometheus metrics for the server and middleware.
+func NewMetricsSubsystem(db *certdb.CertificateRequestsRepository) *PrometheusMetrics {
+	csrs, err := db.RetrieveAll()
```
ghislainbourgeois, line 35: I might be missing something, but doesn't this query get executed only once when setting up the metrics subsystem? It would thus be only able to generate metrics for what exists in the database on startup. I think this might need to be moves inside `generateMetrics`.

internal/metrics/metrics_test.go
```diff
@@ -1,25 +1,39 @@
...
+	"github.com/canonical/gocert/internal/certdb"
 	metrics "github.com/canonical/gocert/internal/metrics"
 )
 
 // TestPrometheusHandler tests that the Prometheus metrics handler responds correctly to an HTTP request.
 func TestPrometheusHandler(t *testing.T) {
-	handler := metrics.NewPrometheusMetricsHandler()
+	db, err := certdb.NewCertificateRequestsRepository(":memory:", "CertificateReq")
+	if err != nil {
+		log.Fatalln(err)
```
ghislainbourgeois, line 26: Shouldn't this be `t.Fatalln(err)`?

internal/metrics/metrics_test.go
```diff
@@ -30,4 +44,112 @@ func TestPrometheusHandler(t *testing.T) {
 	if !strings.Contains(recorder.Body.String(), "go_goroutines") {
 		t.Errorf("handler returned an empty body")
 	}
+	err = db.Close()
+	if err != nil {
+		log.Fatalln(err)
```
ghislainbourgeois, line 49: Shouldn't this be `t.Fatalln(err)`?

internal/metrics/metrics_test.go
```diff
@@ -30,4 +44,112 @@ func TestPrometheusHandler(t *testing.T) {
...
+			log.Fatalf("couldn't create test cert:%s", err)
+		}
+	}
+}
+
+// TestMetrics tests some of the metrics that we currently collect.
+func TestMetrics(t *testing.T) {
+	db, err := certdb.NewCertificateRequestsRepository(":memory:", "CertificateReq")
+	if err != nil {
+		log.Fatalln(err)
```
ghislainbourgeois, line 100: Shouldn't this be `t.Fatalln(err)`?

internal/metrics/metrics_test.go
```diff
@@ -30,4 +44,112 @@ func TestPrometheusHandler(t *testing.T) {
...
+		} else if strings.Contains(line, "certificates_expiring_in_90_days ") && !strings.HasPrefix(line, "#") {
+			if !strings.HasSuffix(line, "3") {
+				t.Errorf("certificates_expiring_in_90_days expected to receive 3")
+			}
+		}
+	}
+
+	err = db.Close()
+	if err != nil {
+		log.Fatalln(err)
```
ghislainbourgeois, line 153: Shouldn't this be `t.Fatalln(err)`?

ghislainbourgeois (changes requested):
Looks good overall, I left some small comments.

internal/metrics/metrics.go
kayra1: good catch, this went through a bunch of iterations and ended up broken

internal/metrics/metrics_test.go
kayra1: yes

internal/metrics/metrics_test.go
kayra1: yes

internal/metrics/metrics_test.go
kayra1: yes

internal/metrics/metrics_test.go
kayra1: yes

internal/metrics/metrics.go
```diff
@@ -1,17 +1,224 @@
...
+	CertificatesExpiringIn7Days    prometheus.Gauge
+	CertificatesExpiringIn30Days   prometheus.Gauge
+	CertificatesExpiringIn90Days   prometheus.Gauge
+	ExpiredCertificates            prometheus.Gauge
+
+	RequestsTotal    prometheus.CounterVec
+	RequestsDuration prometheus.HistogramVec
+}
+
+// NewMetricsSubsystem returns the metrics endpoint HTTP handler and the Prometheus metrics for the server and middleware.
```
saltiyazan, line 33: For me this comment is a bit confusing to only say that we return a `PrometheusMetrics` object.
What do you mean by "and the Prometheus metrics for the server and middleware"?

internal/metrics/metrics.go
```diff
@@ -1,17 +1,224 @@
...
+		CertificatesExpiringIn90Days:   certificatesExpiringIn90DaysMetric(),
+
+		RequestsTotal:    requestsTotalMetric(),
+		RequestsDuration: requestDurationMetric(),
+	}
+	m.registry.MustRegister(m.CertificateRequests)
+	m.registry.MustRegister(m.OutstandingCertificateRequests)
+	m.registry.MustRegister(m.Certificates)
+	m.registry.MustRegister(m.ExpiredCertificates)
+	m.registry.MustRegister(m.CertificatesExpiringIn1Day)
```
saltiyazan, line 73: Why not do something like 
`m.registry.MustRegister(certificatesExpiringIn1DayMetric())` and get rid of these variables like `CertificatesExpiringIn1Day` in `PrometheusMetrics`?

internal/metrics/metrics.go
```diff
@@ -1,17 +1,224 @@
...
+				expiringIn30DaysCertCount += 1
+			}
+			if daysRemaining < 90 {
+				expiringIn90DaysCertCount += 1
+			}
+		}
+	}
+	pm.CertificateRequests.Set(csrCount)
+	pm.OutstandingCertificateRequests.Set(outstandingCSRCount)
+	pm.Certificates.Set(certCount)
```
saltiyazan, line 127: This not a review comment but a question to learn :D
What does this Set do? what is the end result?
Where does the value fit in the following?
```
func certificatesMetric() prometheus.Gauge {
	metric := prometheus.NewGauge(prometheus.GaugeOpts{
		Name: "certificates",
		Help: "Total number of certificates provided to certificate requests",
	})
	return metric
}
```

internal/metrics/metrics.go
kayra1: The metric objects are responsible to keep track of the value of the metric so that they can report it to the collector. In this case, prometheus provides both the metric objects and the collector. The metrics that exist are collected from either the server or the middleware.

internal/metrics/metrics.go
kayra1: We save the metrics objects because we need to access them whenever we want to update the values in them. In this case we don't use them directly because we aren't using it outside of this file, but they are public so that if we have a better way to collect or look at any metrics in the future, we don't need to do any complex logic to inspect it. For example, if we want to show one of the metrics in the front end it would be trivial to just look at this object within a handler.

internal/metrics/metrics.go
kayra1: The name `gauge` comes from a real life [gauge](https://catalog.noshok.com/Asset/60_800_200_PSI-3_4.jpg), It's simply one of three types of metrics that can be collected.

A prom [Gauge](https://prometheus.io/docs/concepts/metric_types/#gauge) is initialized with options that describe what it is, but it is fundamentally a gauge with a discrete number as a value. We can Increment or Decrement this value by 1, but the prom golang client [recommends](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#NewGauge) that you just Set the value directly if you can.

gruyaume (approved):

internal/metrics/metrics.go
saltiyazan: Then how about? If I understood it correctly: 
```suggestion
// NewMetricsSubsystem returns the metrics endpoint HTTP handler and the Prometheus metrics collectors for the server and middleware.
```

internal/metrics/metrics.go
saltiyazan: Alright, thanks for explaining.

saltiyazan (approved):
Looks good. Assuming we resolve [this](https://github.com/canonical/gocert/pull/16#discussion_r1622522579).

ghislainbourgeois (approved):

PR #21, opened 2024-06-13
feat: Certificate Requests Table Query

kayra1:
# Description

e

This PR introduces the query required to get all of the CSR's to dynamically create the table. The frontend will show a very simple loading text when the query fires, a very simple error text when the query fails, and the expected table when the query succeeds. The table is loaded with an array of CSR objects, which the table will pass on to the rows.

There is also the code to ingest the CSR pem string into a native object using `pkijs`, but these functions are incomplete and only return the Common Name and the NotAfter field for now. I've scoped the task to split these out to different tickets named below.

There are some usability improvements that could also be done here, and I've scoped them out into a different ticket too.

Q: Where are the query unit tests?
A: neither vitest nor `react-query` provide facilities to mock queries. Tests for queries will be done in the integration tests that are scoped out in TLSENG-263

## Testing
You may have noticed that there are no tests associated with this PR. This is because there are no react state changes introduced with this PR (no modal popups etc.) and `react-query` does not have unit tests. Any external query tests that need to be written should be through the integration tests, which will come with TLSENG-263

Since the query is written so that it's made to the same endpoint as the server (and to avoid cors errors etc.), the only way to test it is to build it and run gocert with it:
```
npm run build --prefix ui && go install ./...
gocert -config config.yaml
```

`npm run dev` will not work.
If you run into TLS errors, you will need to accept the TLS certificate that gocert sends before the query starts working properly.

## This PR does NOT include:
extracting the values of the CSR into the expanded row: That's TLSENG-276
making it look nicer: That's also TLSENG-276. I also added some recommendations there like making the bg color of certificates that are expired red etc.
Any way to add CSR's from within the frontend: That's TLSENG-265

# Checklist:


CONTRIBUTING.md
```diff
@@ -42,4 +42,4 @@ rockcraft pack -v
 version=$(yq '.version' rockcraft.yaml)
 sudo skopeo --insecure-policy copy oci-archive:gocert_${version}_amd64.rock docker-daemon:gocert:${version}
```
ghislainbourgeois, line 43: Can you please update this also to use the `skopeo` bundled with `rockcraft`?

ghislainbourgeois (approved):
LGTM, I added a small comments, but it could also be done in another PR.

CONTRIBUTING.md
kayra1: done, thanks for the catch

.github/workflows/build-rock.yaml
gruyaume, line 1: This was added in PR #25 (the same applies to a couple of related changes here)

ui/src/app/certificate_requests/page.tsx
gruyaume, line 1: This is probably not related to this PR.

The empty table looks odd. Until there are Certificate Requests to show in the table, we likely want an "empty state" page so that users are directed to generate a CSR with a big green button front and center. Ex. 

> No Certificate Request Available
> [Generate]

## Reference
- https://vanillaframework.io/docs/patterns/empty-state

gruyaume (commented):
Looks good to me besides the two comments

.github/workflows/build-rock.yaml
kayra1: merged main

ui/src/app/certificate_requests/page.tsx
kayra1: Added an empty state page where the Add New CSR button is moved to the center.

PR #31, opened 2024-06-21
feat: Extract and show csr details

saltiyazan:
# Description

- Extract more details from the CSR
- Show the extracted details
- Show `N/A` if a value is missing
- Minor refactor in the code to improve the decoding of the PEM string

Screenshot 2024-06-24 at 11 51 49

# Checklist:


ui/src/app/certificate_requests/row.tsx
```diff
@@ -67,7 +67,14 @@ export default function Row({ id, csr, certificate, ActionMenuExpanded, setActio
...
                 <div className="row">
                     <div className="col-8">
-                        <p><b>Common Name</b>: {csrObj.subjects.find((e) => e.type == "Common Name")?.value}</p>
+                        <p><b>Common Name</b>: {csrObj.commonName || "N/A"}</p>
+                        <p><b>Organization</b>: {csrObj.organization || "N/A"}</p>
+                        <p><b>Country</b>: {csrObj.country || "N/A"}</p>
+                        <p><b>Locality</b>: {csrObj.locality || "N/A"}</p>
+                        <p><b>Email Address</b>: {csrObj.emailAddress || "N/A"}</p>
+                        <p><b>Subject Alternative Name DNS</b>: {(csrObj.sansDns && csrObj.sansDns.length > 0) ? csrObj.sansDns.join(', ') : "N/A"}</p>
+                        <p><b>Subject Alternative Name IP addresses</b>: {(csrObj.sansIp && csrObj.sansIp.length > 0) ? csrObj.sansIp.join(', ') : "N/A"}</p>
```
kayra1, line 76: lets not have the paranthesis in the beginning of the tertiary operator to be consistent with the rest of the code

ui/src/app/nav.tsx
```diff
@@ -18,7 +18,11 @@ function SubmitCSR({ csrText, onClickFunc }: { csrText: string, onClickFunc: any
         extractCSR(csrText.trim())
         csrIsValid = true
     }
-    catch { }
+    catch (e) {
+        if (e instanceof Error) {
+            console.log(e.message)
+        }
+    }
```
kayra1, line 25: There is no point in logging this, it will just spam the console with the same generic `couldn't process csr` error messages since this runs every time the text changes. We already return a different component when it fails.

ui/src/app/utils.ts
```diff
@@ -38,62 +39,107 @@ export const oidToName = (oid: string) => {
...
+        emailAddress,
+        country,
+        locality,
+        sansDns,
+        sansIp,
+        is_ca
+    }
 }
 
 export const extractCert = (certPemString: string) => {
```
kayra1, line 135: Please create a new ticket and change the description of 276 if you aren't going to tackle extracting certificate information here.

kayra1 (changes requested):
mostly LGTM

ui/src/app/utils.ts
saltiyazan: Ticket created

ui/src/app/certificate_requests/row.tsx
saltiyazan: ✅

ui/src/app/nav.tsx
saltiyazan: ✅

kayra1 (approved):
Tentatively approved

ui/src/app/certificate_requests/row.tsx
```diff
@@ -67,7 +67,14 @@ export default function Row({ id, csr, certificate, ActionMenuExpanded, setActio
...
                     <div className="col-8">
-                        <p><b>Common Name</b>: {csrObj.subjects.find((e) => e.type == "Common Name")?.value}</p>
+                        <p><b>Common Name</b>: {csrObj.commonName || "N/A"}</p>
+                        <p><b>Organization</b>: {csrObj.organization || "N/A"}</p>
+                        <p><b>Country</b>: {csrObj.country || "N/A"}</p>
+                        <p><b>Locality</b>: {csrObj.locality || "N/A"}</p>
+                        <p><b>Email Address</b>: {csrObj.emailAddress || "N/A"}</p>
+                        <p><b>Subject Alternative Name DNS</b>: {csrObj.sansDns && csrObj.sansDns.length > 0 ? csrObj.sansDns.join(', ') : "N/A"}</p>
+                        <p><b>Subject Alternative Name IP addresses</b>: {csrObj.sansIp && csrObj.sansIp.length > 0 ? csrObj.sansIp.join(', ') : "N/A"}</p>
+                        <p><b>Certificate request for a certificate authority</b>: {csrObj.is_ca ? "Yes" : "No"}</p>
```
gruyaume, line 77: Currently there is a lot of `N/A`'s showing up cluttering vertical space and it's not obvious to the eye where to look. Here are a couple of ideas: 

1. Have the key (whatever the cert field is) and Value (`N/A`) be grey instead of black for cases where the value is empty. This would make it clearer where to look.
2. Don't show the cert key/value at all when they are empty. This would reduce the number of unnecessary vertical space.

My recommended approach would be to go with option 2. In either case I would recommend experimenting with these 2 and any other idea you may have. 

image

ui/src/app/certificate_requests/row.tsx
gruyaume, line 1: Should we also include cert field retrieval in the "New Certificate Request" Aside?

gruyaume (changes requested):

ui/src/app/certificate_requests/row.tsx
kayra1: we could, but probably should make it a separate ticket

ui/src/app/certificate_requests/row.tsx
kayra1: It would also be nice to reduce the margins between these fields to have it be a bit more compact.

kayra1: I like seeing the N/A values because I usually forget what fields a CSR or a Cert could have, and it could be useful for users to quickly identify if they forgot to fill out a field if they don't have the fields memorized. Since i'm not the target audience for this, im open to having it either way.

ui/src/app/certificate_requests/row.tsx
saltiyazan: Yes a ticket was created in the backlog for certificate details.

ui/src/app/certificate_requests/row.tsx
saltiyazan: I pushed some changes greying out missing info and reducing the margins.
Here is how it looks now.

saltiyazan: I experimented with not displaying the missing info. And yes it would like nicer.
But I believe having the N/A values for missing fields provides good useful information as it would tell the user that those values were not provided, which is not necessary a trivial thing to get from not showing the fields.
Here is how it would look.
Screenshot 2024-06-26 at 11 36 12

ui/src/app/certificate_requests/row.tsx
```diff
@@ -100,9 +106,20 @@ export default function Row({ id, csr, certificate, ActionMenuExpanded, setActio
...
+                            <h4>Basic Information</h4>
+                            {getFieldDisplay("Common Name", csrObj.commonName, "N/A")}
+                            {getFieldDisplay("Organization", csrObj.organization, "N/A")}
+                            {getFieldDisplay("Country", csrObj.country, "N/A")}
+                            {getFieldDisplay("Locality", csrObj.locality, "N/A")}
+                            {getFieldDisplay("Email Address", csrObj.emailAddress, "N/A")}
+                            <p><b>Certificate request for a certificate authority</b>: {csrObj.is_ca ? "Yes" : "No"}</p>
+                        </div>
+                        <div className="certificate-info">
+                            <h4>Subject Alternative Names</h4>
```
gruyaume, line 120: I don't know about separating the fields into those arbitrary categories (Basic Information and Subject Alternative Names). For now I would recommend going with showing the fields in an order that makes sense (ex. having subject alternative names should likely be quite high, right after common name). 

In the future we can change this based on user feedback if necessary

gruyaume (changes requested):

ui/src/app/certificate_requests/row.tsx
saltiyazan: I still think it's valuable information. But I don't have a better argument for now, so I changed it.
Done.

ui/src/app/certificate_requests/row.tsx
saltiyazan: Removed and re-ordered.

ui/src/app/certificate_requests/row.tsx
```diff
@@ -100,9 +109,17 @@ export default function Row({ id, csr, certificate, ActionMenuExpanded, setActio
                     </span>
                 </td>
                 <td id="expanded-row" className="p-table__expanding-panel" aria-hidden={detailsMenuOpen ? "false" : "true"}>
-                    <div className="row">
-                        <div className="col-8">
-                            <p><b>Common Name</b>: {csrObj.subjects.find((e) => e.type == "Common Name")?.value}</p>
+                    <div className="col-8">
+                        <div className="certificate-info">
+                            <h4>CSR Details</h4>
```
gruyaume, line 114:
```suggestion
```

gruyaume (changes requested):
One last detail, the rest looks good

ui/src/app/certificate_requests/row.tsx
```diff
@@ -100,9 +109,17 @@ export default function Row({ id, csr, certificate, ActionMenuExpanded, setActio
...
+                            {getFieldDisplay("Common Name", csrObj.commonName)}
+                            {getFieldDisplay("Subject Alternative Name DNS", csrObj.sansDns && csrObj.sansDns.length > 0 ? csrObj.sansDns.join(', ') : "")}
+                            {getFieldDisplay("Subject Alternative Name IP addresses", csrObj.sansIp && csrObj.sansIp.length > 0 ? csrObj.sansIp.join(', ') : "")}
+                            {getFieldDisplay("Organization", csrObj.organization)}
+                            {getFieldDisplay("Country", csrObj.country)}
+                            {getFieldDisplay("Locality", csrObj.locality)}
```
gruyaume, line 120: We have "Country" and "Locality" but not "State or Province", I think we should add this one as well

ui/src/app/certificate_requests/row.tsx
```diff
@@ -100,9 +109,17 @@ export default function Row({ id, csr, certificate, ActionMenuExpanded, setActio
...
-                    <div className="row">
-                        <div className="col-8">
-                            <p><b>Common Name</b>: {csrObj.subjects.find((e) => e.type == "Common Name")?.value}</p>
+                    <div className="col-8">
+                        <div className="certificate-info">
+                            <h4>CSR Details</h4>
+                            {getFieldDisplay("Common Name", csrObj.commonName)}
+                            {getFieldDisplay("Subject Alternative Name DNS", csrObj.sansDns && csrObj.sansDns.length > 0 ? csrObj.sansDns.join(', ') : "")}
+                            {getFieldDisplay("Subject Alternative Name IP addresses", csrObj.sansIp && csrObj.sansIp.length > 0 ? csrObj.sansIp.join(', ') : "")}
+                            {getFieldDisplay("Organization", csrObj.organization)}
```
gruyaume, line 118: In addition to "Organization", we should likely have "Organizational Unit"

gruyaume (changes requested):
My bad for not catching this earlier, those fields should be displayed (at least when their value is not empty)

- Country
- State
- Locality
- Organization
- Organizational Unit
- Common Name
- Alternative Names
- Email Address

Those are the fields you get to choose when using OpenSSL and right now we are missing "State" and "Organizational Unit"

ui/src/app/certificate_requests/row.tsx
saltiyazan: State or Province added.

ui/src/app/certificate_requests/row.tsx
saltiyazan: Organizational Unit added.

ui/src/app/certificate_requests/row.tsx
saltiyazan: Modified.
Screenshot 2024-06-27 at 14 49 42

gruyaume (approved):

PR #23, opened 2024-06-14
feat: Reject & Delete CSR

kayra1:
# Description

This change allows users to reject or delete CSR's by utilizing the two buttons within the action menu. There is a modal popup for each button that asks the user to confirm the action.

Screenshot from 2024-06-21 16-27-01

# Checklist:

- [x] My code follows the [style guidelines](/CONTRIBUTING.md) of this project
- [x] I have performed a self-review of my own code
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that validate the behaviour of the software
- [x] I validated that new and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #32, opened 2024-06-24
feat: Copy & Download CSR

kayra1:
# Description

This PR adds the ability to copy or download CSR's. When copied, a notification indicating a successful copy pops out under the row, and the downloaded file is automatically named `csr.pem`.

depends on PR #23 

# Checklist:


ui/src/app/certificate_requests/row.tsx
gruyaume, line 1: I'm not a fan of the success notification expanding the table and showing in what feels like a random place.

I believe it would make more sense to render it to the right of the action button.

## Current place
image

ui/src/app/certificate_requests/row.tsx
```diff
@@ -47,6 +48,26 @@ export default function Row({ id, csr, certificate, ActionMenuExpanded, setActio
...
+            }, 2500);
+        }, function (err) {
+            console.error('could not copy text: ', err);
+        });
+    }
+    const handleDownload = () => {
+        const blob = new Blob([csr], { type: 'text/plain' });
+        const link = document.createElement('a');
+        link.href = URL.createObjectURL(blob);
+        link.download = "csr.pem"; // TODO: change this to <csr-commonname>.pem
```
gruyaume, line 65: We should probably suffix this with the CSR ID (or with the common name as mentioned in this todo). When we download more than one they all have the same file name

gruyaume (changes requested):
Overall this looks good, I got comments related to the placement of the notification and the fact that all csr files have the same name when downloaded

ui/src/app/certificate_requests/row.tsx
kayra1: I'll append ID to the name of the file, and keep the todo as is, since it's waiting on #31

ui/src/app/certificate_requests/row.tsx
kayra1: moved it there

saltiyazan (approved):
Looks good, given the file name is fixed when #31 is merged.

ui/src/app/certificate_requests/row.tsx
gruyaume: Nice, thank you

gruyaume (approved):

PR #33, opened 2024-06-24
feat: Upload & Revoke Certificate

kayra1:
# Description

This PR introduces the final 2 buttons in the row, uploading and revoking certificates. The upload button shows a modal that's almost identical to the submit CSR form. The revoke button shows a confirmation modal. I've bundled these together into this PR because they are related features and not too large since their underlying logic has already been written.

Depends on #32 

# Checklist:


ui/src/app/certificate_requests/components.tsx
gruyaume, line 1: If i don't remove the newline I'm getting an "Invalid Certificate" Notification

ui/src/app/certificate_requests/components.tsx
```diff
@@ -28,6 +31,88 @@ export function ConfirmationModal({ modalData, setModalData }: ConfirmationModal
...
+                </header>
+                <form className="p-form p-form--stacked">
+                    <div className="p-form__group row">
+                        <label htmlFor="textarea">
+                            Enter or upload the Certificate in PEM format below
+                        </label>
+                        <textarea id="csr-textarea" name="textarea" rows={10} placeholder="-----BEGIN CERTIFICATE-----" onChange={handleTextChange} value={certificatePEMString} />
+                    </div>
+                    <div className="p-form__group row">
+                        <input type="file" name="upload" accept=".pem" onChange={handleFileChange}></input>
```
gruyaume, line 102: It is quite common to see `.crt` too

ui/src/app/certificate_requests/components.tsx
gruyaume, line 1: If we Upload a certificate that was already uploaded we should likely grey out the submit button and mention to the user that "This certificate was already uploaded for this request" or something along those lines

gruyaume (changes requested):

ui/src/app/certificate_requests/components.tsx
kayra1: added

ui/src/app/certificate_requests/components.tsx
kayra1: should be fixed now

kayra1: added a custom message

gruyaume: I still see the button as being enabled
image

gruyaume: Actually it depends on whether or not I add newlines, we should probably strip all of those

kayra1: ah, forgot one more .trim() call :D

ui/src/app/certificate_requests/components.tsx
```diff
@@ -28,6 +31,95 @@ export function ConfirmationModal({ modalData, setModalData }: ConfirmationModal
...
 
+function SubmitCertificate({ existingCSRText, existingCertText, certText, onClickFunc }: { existingCSRText: string, existingCertText: string, certText: string, onClickFunc: any }) {
+    let certIsValid = false
+    try {
+        extractCert(certText.trim())
+        if (csrMatchesCertificate(existingCSRText.trim(), certText.trim())) {
+            certIsValid = true
+        }
+    }
+    catch { }
```
gruyaume, line 42: Here we see "Invalid Certificates" when the certificate is valid but does not match the CSR. I believe we should differentiate those 2 situations with independent error messages.
- The Certificate is invalid
- The Certificate does not match the request

gruyaume (changes requested):

ui/src/app/certificate_requests/components.tsx
kayra1: done

gruyaume (approved):

PR #35, opened 2024-06-28
feat: Extract certificate details

saltiyazan:
# Description

- Improves the code to extract certificates
- Improves the code to compare certificates
- Display certificate details

Both certificate and csr details are displayed, as there is no guarantee that those are equal.
Discrepancies between csr and cert fields are highlighted in red.
The details are responsive and they go into one column on small screens.
The expiry date is coloured green, it turns red when we are within one day into expiry and beyond.

Screenshot 2024-07-01 at 13 40 46

Screenshot 2024-07-01 at 13 40 55

Screenshot 2024-07-01 at 13 39 05

# Checklist:

### DIFF
--- a/ui/src/app/certificate_requests/row.tsx
+++ b/ui/src/app/certificate_requests/row.tsx
@@ -19,6 +19,9 @@ export type ConfirmationModalData = {
 } | null
 
 export default function Row({ id, csr, certificate, ActionMenuExpanded, setActionMenuExpanded }: rowProps) {
+    const red = "#c7162b"
+    const green = "#0e8420"
+    const yellow = "#f99b11"
     const [successNotification, setSuccessNotification] = useState<string | null>(null)
     const [detailsMenuOpen, setDetailsMenuOpen] = useState<boolean>(false)
     const [certificateFormOpen, setCertificateFormOpen] = useState<boolean>(false)
@@ -94,7 +97,7 @@ export default function Row({ id, csr, certificate, ActionMenuExpanded, setActio
         return field ? (
             <p>
                 <b>{label}:</b>{" "}
-                <span style={{ color: isMismatched ? '#c7162b' : 'inherit' }}>
+                <span style={{ color: isMismatched ? red : 'inherit' }}>
                     {field}
                 </span>
             </p>
@@ -110,11 +113,11 @@ export default function Row({ id, csr, certificate, ActionMenuExpanded, setActio
         const timeDifference = expiryDate.getTime() - now.getTime();
 
         if (timeDifference < 0) {
-            return '#c7162b';
+            return red;
         } else if (timeDifference < oneDayInMillis) {
-            return '#f99b11';
+            return yellow;
         } else {
-            return '#0e8420';
+            return green;
         }
     };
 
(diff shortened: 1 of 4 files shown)


ui/src/app/certificate_requests/row.tsx
```diff
@@ -4,6 +4,7 @@ import { extractCSR, extractCert } from "../utils"
 import { deleteCSR, rejectCSR, revokeCertificate } from "../queries"
 import { ConfirmationModal, SubmitCertificateModal, SuccessNotification } from "./components"
 import "./../globals.scss"
+import { get } from "http"
```
kayra1, line 7: is this supposed to be here?

ui/src/app/certificate_requests/row.tsx
```diff
@@ -88,13 +89,33 @@ export default function Row({ id, csr, certificate, ActionMenuExpanded, setActio
...
+
+        const expiryDate = new Date(notAfter);
+        const now = new Date();
+        const oneDayInMillis = 24 * 60 * 60 * 1000;
+        const timeDifference = expiryDate.getTime() - now.getTime();
+
+        if (timeDifference < 0 || timeDifference < oneDayInMillis) {
+            return 'red';
+        } else {
+            return 'green';
```
kayra1, line 116: how about doing yellow instead of red for the ones that are expiring in 24 hours? Also how about highlighting them in that color instead of changing the font color? Also we should use the same colors as the [semantic](https://vanillaframework.io/docs/settings/color-settings) colors everywhere else:
green: #0e8420
yellow: #f99b11
red: #c7162b

ui/src/app/certificate_requests/row.tsx
```diff
@@ -145,21 +166,44 @@ export default function Row({ id, csr, certificate, ActionMenuExpanded, setActio
...
+                                    {getFieldDisplay("Email Address", certObj.emailAddress, csrObj.emailAddress)}
+                                    {getFieldDisplay("Start of validity", certObj.notBefore)}
+                                    {getFieldDisplay("Expiry Time", certObj.notAfter)}
+                                    {getFieldDisplay("Issuer Common Name", certObj.issuerCommonName)}
+                                </div>
+                            </div>
+                        )}
                     </div>
                 </td>
+
```
kayra1, line 206: hahaha
```suggestion
```

ui/src/app/utils.ts
```diff
@@ -136,76 +151,88 @@ export const extractCSR = (csrPemString: string) => {
...
     const binaryDer = pemToArrayBuffer(certPemString);
-
-    // Parse DER encoded certificate
     const asn1 = fromBER(binaryDer);
     if (asn1.offset === -1) {
         throw new Error("Error parsing certificate");
     }
 
-    // Load Certificate object
     const cert = new Certificate({ schema: asn1.result });
```
kayra1, line 166: why not `const cert = loadCertificate(certPemString)`? You've already written the function

kayra1 (changes requested):
I really like how it looks. I especially like that if the common name is very long the certificate details wrap.

There are still some suggestions though.

ui/src/app/utils.ts
saltiyazan: Yup, done. Thanks

ui/src/app/certificate_requests/row.tsx
saltiyazan: Fixed, thanks

kayra1: I would write `const green = #c7162b` etc. at the top to make it more readable.

saltiyazan: done

kayra1 (approved):
I think we can pass on the highlighting for now.

kayra1 (approved):

PR #36, opened 2024-07-01
feat: User table SQL queries

kayra1:
# Description

This change introduces the SQL queries required for the GoCert authorization flow as defined in TE091. The password salting and hashing algorithm required for storage is also done at the database level.

# Checklist:


ui/src/app/types.ts
saltiyazan, line 1: Not like it is super important, but why did you change the names here?

saltiyazan (approved):

ui/src/app/types.ts
kayra1: When I first wrote the handler I wasn't sure how to map the struct to different json key names. I know now, so I changed it to be consistent with all the other json objects we have.

PR #37, opened 2024-07-02
feat: Handlers to create and fetch accounts

saltiyazan:
# Description

- Adds Get and Get by ID handlers
- Adds Post handler
- Adds tests
- Improves response codes when resource is not found

# Checklist:

### DIFF
--- a/internal/api/handlers.go
+++ b/internal/api/handlers.go
@@ -265,9 +265,9 @@ func GetUserAccounts(env *Environment) http.HandlerFunc {
 			logErrorAndWriteResponse(err.Error(), http.StatusInternalServerError, w)
 			return
 		}
-		apiUsers := make([]certdb.UserAPIModel, 0, len(users))
+		apiUsers := make([]certdb.User, 0, len(users))
 		for _, user := range users {
-			apiUsers = append(apiUsers, certdb.UserAPIModel{
+			apiUsers = append(apiUsers, certdb.User{
 				ID:          user.ID,
 				Username:    user.Username,
 				Permissions: user.Permissions,
@@ -298,18 +298,13 @@ func GetUserAccount(env *Environment) http.HandlerFunc {
 			logErrorAndWriteResponse(err.Error(), http.StatusInternalServerError, w)
 			return
 		}
-		apiUser := certdb.UserAPIModel{
+		apiUser := certdb.User{
 			ID:          userAccount.ID,
 			Username:    userAccount.Username,
 			Permissions: userAccount.Permissions,
 		}
 
 		body, err := json.Marshal(apiUser)
-		if err != nil {
-			logErrorAndWriteResponse(err.Error(), http.StatusInternalServerError, w)
-			return
-		}
-
 		if err != nil {
 			logErrorAndWriteResponse(err.Error(), http.StatusInternalServerError, w)
 			return
@@ -323,13 +318,13 @@ func GetUserAccount(env *Environment) http.HandlerFunc {
 // PostUserAccount creates a new User Account, and returns the id of the created row
 func PostUserAccount(env *Environment) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
-		var user certdb.UserCreationRequest
+		var user certdb.User
 		if err := json.NewDecoder(r.Body).Decode(&user); err != nil {
 			logErrorAndWriteResponse("Invalid JSON format", http.StatusBadRequest, w)
 			return
 		}
-		if user.Username == "" || user.Password == "" {
-			logErrorAndWriteResponse("Username and password are required", http.StatusBadRequest, w)
+		if user.Username == "" {
+			logErrorAndWriteResponse("Username is required", http.StatusBadRequest, w)
 			return
 		}
 		users, err := env.DB.RetrieveAllUsers()
--- a/internal/api/handlers.go
+++ b/internal/api/handlers.go
@@ -28,7 +28,7 @@ func NewGoCertRouter(env *Environment) http.Handler {
 	apiV1Router.HandleFunc("POST /certificate_requests/{id}/certificate/reject", RejectCertificate(env))
 	apiV1Router.HandleFunc("DELETE /certificate_requests/{id}/certificate", DeleteCertificate(env))
 
-	apiV1Router.HandleFunc("GET /accounts/{id}/account", GetUserAccount(env))
+	apiV1Router.HandleFunc("GET /accounts/{id}", GetUserAccount(env))
 	apiV1Router.HandleFunc("GET /accounts", GetUserAccounts(env))
 	apiV1Router.HandleFunc("POST /accounts", PostUserAccount(env))
 
@@ -265,15 +265,10 @@ func GetUserAccounts(env *Environment) http.HandlerFunc {
 			logErrorAndWriteResponse(err.Error(), http.StatusInternalServerError, w)
 			return
 		}
-		apiUsers := make([]certdb.User, 0, len(users))
-		for _, user := range users {
-			apiUsers = append(apiUsers, certdb.User{
-				ID:          user.ID,
-				Username:    user.Username,
-				Permissions: user.Permissions,
-			})
+		for i := range users {
+			users[i].Password = ""
 		}
-		body, err := json.Marshal(apiUsers)
+		body, err := json.Marshal(users)
 		if err != nil {
 			logErrorAndWriteResponse(err.Error(), http.StatusInternalServerError, w)
 			return
@@ -298,13 +293,8 @@ func GetUserAccount(env *Environment) http.HandlerFunc {
 			logErrorAndWriteResponse(err.Error(), http.StatusInternalServerError, w)
 			return
 		}
-		apiUser := certdb.User{
-			ID:          userAccount.ID,
-			Username:    userAccount.Username,
-			Permissions: userAccount.Permissions,
-		}
-
-		body, err := json.Marshal(apiUser)
+		userAccount.Password = ""
+		body, err := json.Marshal(userAccount)
 		if err != nil {
 			logErrorAndWriteResponse(err.Error(), http.StatusInternalServerError, w)
 			return
--- a/internal/api/handlers.go
+++ b/internal/api/handlers.go
@@ -1,12 +1,14 @@
 package server
 
 import (
+	"crypto/rand"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
 	"io/fs"
 	"log"
+	"math/big"
 	"net/http"
 	"strconv"
 	"strings"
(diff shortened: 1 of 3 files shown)


internal/api/handlers.go
```diff
@@ -252,6 +257,110 @@ func DeleteCertificate(env *Environment) http.HandlerFunc {
...
+			return
+		}
+		apiUsers := make([]certdb.UserAPIModel, 0, len(users))
+		for _, user := range users {
+			apiUsers = append(apiUsers, certdb.UserAPIModel{
+				ID:          user.ID,
+				Username:    user.Username,
+				Permissions: user.Permissions,
+			})
+		}
```
kayra1, line 275: ideally you can just modify the tags in the struct to automatically marshall the output correctly and not have to do it manually here.

internal/api/handlers.go
```diff
@@ -125,7 +130,7 @@ func GetCertificateRequest(env *Environment) http.HandlerFunc {
 		cert, err := env.DB.RetrieveCSR(id)
 		if err != nil {
 			if err.Error() == "csr id not found" {
-				logErrorAndWriteResponse(err.Error(), http.StatusBadRequest, w)
+				logErrorAndWriteResponse(err.Error(), http.StatusNotFound, w)
```
kayra1, line 133: I'm still not sure if we should use a 404 here. It may indicate to users that the whole URL is wrong instead of just the resource being missing. But at the same time a lot of websites do exactly that.

kayra1 (commented):
Definitely going in the right direction

internal/api/handlers.go
```diff
@@ -252,6 +257,110 @@ func DeleteCertificate(env *Environment) http.HandlerFunc {
...
+		body, err := json.Marshal(apiUser)
+		if err != nil {
+			logErrorAndWriteResponse(err.Error(), http.StatusInternalServerError, w)
+			return
+		}
+
+		if err != nil {
+			logErrorAndWriteResponse(err.Error(), http.StatusInternalServerError, w)
+			return
+		}
```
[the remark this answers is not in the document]
saltiyazan: Yup they were

internal/api/handlers_test.go
```diff
@@ -328,6 +334,54 @@ func TestGoCertRouter(t *testing.T) {
...
+			status:   http.StatusNotFound,
+		},
+		{
+			desc:     "Create second user success",
+			method:   "POST",
+			path:     "/api/v1/accounts",
+			data:     invalidUser,
+			response: "error: Username and password are required",
+			status:   http.StatusBadRequest,
+		},
```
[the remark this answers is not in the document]
saltiyazan: Done

internal/certdb/certdb.go
```diff
@@ -57,6 +57,17 @@ type User struct {
...
+type UserCreationRequest struct {
+	Username string `json:"username"`
+	Password string `json:"password"`
+}
+
+type UserAPIModel struct {
+	ID          int    `json:"id"`
+	Username    string `json:"username"`
+	Permissions int    `json:"permissions"`
+}
```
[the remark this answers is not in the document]
saltiyazan: Done

internal/api/handlers.go
```diff
@@ -252,6 +257,110 @@ func DeleteCertificate(env *Environment) http.HandlerFunc {
...
+	return func(w http.ResponseWriter, r *http.Request) {
+		var user certdb.UserCreationRequest
+		if err := json.NewDecoder(r.Body).Decode(&user); err != nil {
+			logErrorAndWriteResponse("Invalid JSON format", http.StatusBadRequest, w)
+			return
+		}
+		if user.Username == "" || user.Password == "" {
+			logErrorAndWriteResponse("Username and password are required", http.StatusBadRequest, w)
+			return
+		}
```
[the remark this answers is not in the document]

internal/api/handlers.go

internal/api/handlers.go
saltiyazan: Yeah, I tried to do some research and 404 seemed the most appropriate for me, plus the error message `user id not found` is clear I think.

internal/api/handlers.go
```diff
@@ -252,6 +257,110 @@ func DeleteCertificate(env *Environment) http.HandlerFunc {
...
+
+		permission := "0"
+		if len(users) == 0 {
+			permission = "1" //if this is the first user it will be admin
+		}
+		id, err := env.DB.CreateUser(user.Username, user.Password, permission)
+		if err != nil {
+			if strings.Contains(err.Error(), "UNIQUE constraint failed") {
```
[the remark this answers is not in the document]
kayra1: eh, lets ignore it there's a bunch of error cleanup that i need to do here in a different pr

internal/api/handlers.go
kayra1: sure lets keep it

internal/api/handlers.go
```diff
@@ -27,6 +28,10 @@ func NewGoCertRouter(env *Environment) http.Handler {
 	apiV1Router.HandleFunc("POST /certificate_requests/{id}/certificate/reject", RejectCertificate(env))
 	apiV1Router.HandleFunc("DELETE /certificate_requests/{id}/certificate", DeleteCertificate(env))
 
+	apiV1Router.HandleFunc("GET /accounts/{id}/account", GetUserAccount(env))
```
kayra1, line 31:
```suggestion
	apiV1Router.HandleFunc("GET /accounts/{id}", GetUserAccount(env))
```

internal/api/handlers.go
```diff
@@ -252,6 +257,105 @@ func DeleteCertificate(env *Environment) http.HandlerFunc {
...
+		}
+		apiUsers := make([]certdb.User, 0, len(users))
+		for _, user := range users {
+			apiUsers = append(apiUsers, certdb.User{
+				ID:          user.ID,
+				Username:    user.Username,
+				Permissions: user.Permissions,
+			})
+		}
+		body, err := json.Marshal(apiUsers)
```
kayra1, line 276: This is why we modified the json tags on the structs. Read below on why we set the password to `""`
```suggestion
		for i := range users {
			users[i].Password = ""
		}
		body, err := json.Marshal(users)
```

internal/api/handlers.go
```diff
@@ -252,6 +257,105 @@ func DeleteCertificate(env *Environment) http.HandlerFunc {
...
+			logErrorAndWriteResponse(err.Error(), http.StatusInternalServerError, w)
+			return
+		}
+		apiUser := certdb.User{
+			ID:          userAccount.ID,
+			Username:    userAccount.Username,
+			Permissions: userAccount.Permissions,
+		}
+
+		body, err := json.Marshal(apiUser)
```
kayra1, line 307: This is why we modified the json tags on the structs. Read below on why we set the password to `""`
```suggestion
		userAccount.Password = ""
		body, err := json.Marshal(userAccount)
```

internal/certdb/certdb.go
```diff
@@ -53,8 +53,8 @@ type CertificateRequest struct {
 type User struct {
 	ID          int    `json:"id"`
 	Username    string `json:"username"`
-	Password    string `json:"password"`
-	Permissions int    `json:"permissions"`
+	Password    string `json:"-"`
+	Permissions *int   `json:"permissions"`
```
kayra1, line 57: a `-` will always remove the field. `omitempty` will only remove the field when the value is a nil value. So if we set the password to `""` above, we can remove it from the json string we marshalled, but still keep the ability to look at the password (we will need to check it when logging in)
```suggestion
	Password    string `json:"password,omitempty"`
	Permissions *int   `json:"permissions"`
```

kayra1 (changes requested):
there is an easier way to deal with the json

internal/api/handlers.go
saltiyazan: ✅

internal/api/handlers.go
saltiyazan: ✅

internal/api/handlers.go
saltiyazan: ✅

internal/certdb/certdb.go
saltiyazan: ✅

internal/api/handlers.go
```diff
@@ -252,6 +257,95 @@ func DeleteCertificate(env *Environment) http.HandlerFunc {
...
+	return func(w http.ResponseWriter, r *http.Request) {
+		var user certdb.User
+		if err := json.NewDecoder(r.Body).Decode(&user); err != nil {
+			logErrorAndWriteResponse("Invalid JSON format", http.StatusBadRequest, w)
+			return
+		}
+		if user.Username == "" {
+			logErrorAndWriteResponse("Username is required", http.StatusBadRequest, w)
+			return
+		}
```
kayra1, line 319: we should create a random password here if the user.Password is empty.

internal/api/handlers.go
```diff
@@ -252,6 +257,95 @@ func DeleteCertificate(env *Environment) http.HandlerFunc {
...
+				logErrorAndWriteResponse("user with given username already exists", http.StatusBadRequest, w)
+				return
+			}
+			logErrorAndWriteResponse(err.Error(), http.StatusInternalServerError, w)
+			return
+		}
+
+		w.Header().Set("Content-Type", "application/json")
+		w.WriteHeader(http.StatusCreated)
+		response := fmt.Sprintf(`{"id": %d}`, id)
```
kayra1, line 342: we should write `{"password": %s"}, user.Password` alongside the id if we generated a new password.

kayra1 (changes requested):
almost done

internal/api/handlers.go
saltiyazan: Done

internal/api/handlers.go

internal/api/handlers.go
```diff
@@ -356,3 +364,16 @@ func logErrorAndWriteResponse(msg string, status int, w http.ResponseWriter) {
 		logErrorAndWriteResponse(err.Error(), http.StatusInternalServerError, w)
 	}
 }
+
+var GeneratePassword = func(length int) (string, error) {
```
kayra1, line 368: We could instead of returning a (string, error), return just a string and return `""` when an error occurs. This is kinda subjective.

internal/api/handlers.go
```diff
@@ -318,6 +320,9 @@ func PostUserAccount(env *Environment) http.HandlerFunc {
 			logErrorAndWriteResponse("Username is required", http.StatusBadRequest, w)
 			return
 		}
+		if user.Password == "" {
+			user.Password, _ = GeneratePassword(8)
```
kayra1, line 324: If the generatepassword fails, maybe we can just give a default password like `password123` instead of giving a 500 server error.

saltiyazan: As discussed let's stick to the error and not have a hardcoded password.

kayra1 (approved):
great work, thanks

internal/certdb/certdb.go
```diff
@@ -53,10 +53,12 @@ type CertificateRequest struct {
 type User struct {
 	ID          int    `json:"id"`
 	Username    string `json:"username"`
-	Password    string `json:"password"`
+	Password    string `json:"password,omitempty"`
 	Permissions int    `json:"permissions"`
 }
 
+var ErrIdNotFound = errors.New("id not found")
```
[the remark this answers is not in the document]
saltiyazan: The error type is not compile time constant. I get `errors.New("id not found") (value of type error) is not constantcompiler`

PR #40, opened 2024-07-04
feat: Adds login endpoint and handlers

saltiyazan:
# Description

- Adds login endpoint and handlers
- Adds test cases
- Generates a new jwt secret on server restarts

# Checklist:

### DIFF
--- a/internal/api/handlers.go
+++ b/internal/api/handlers.go
@@ -12,10 +12,13 @@ import (
 	"net/http"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/canonical/gocert/internal/certdb"
 	metrics "github.com/canonical/gocert/internal/metrics"
 	"github.com/canonical/gocert/ui"
+	"github.com/golang-jwt/jwt"
+	"golang.org/x/crypto/bcrypt"
 )
 
 // NewGoCertRouter takes in an environment struct, passes it along to any handlers that will need
@@ -34,6 +37,8 @@ func NewGoCertRouter(env *Environment) http.Handler {
 	apiV1Router.HandleFunc("GET /accounts/{id}", GetUserAccount(env))
 	apiV1Router.HandleFunc("GET /accounts", GetUserAccounts(env))
 	apiV1Router.HandleFunc("POST /accounts", PostUserAccount(env))
+
+	apiV1Router.HandleFunc("POST /login", Login(env))
 
 	m := metrics.NewMetricsSubsystem(env.DB)
 	frontendHandler := newFrontendFileServer()
@@ -360,6 +365,52 @@ func PostUserAccount(env *Environment) http.HandlerFunc {
 	}
 }
 
+func Login(env *Environment) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		var userRequest certdb.User
+		if err := json.NewDecoder(r.Body).Decode(&userRequest); err != nil {
+			logErrorAndWriteResponse("Invalid JSON format", http.StatusBadRequest, w)
+			return
+		}
+		if userRequest.Username == "" {
+			logErrorAndWriteResponse("Username is required", http.StatusBadRequest, w)
+			return
+		}
+		if userRequest.Password == "" {
+			logErrorAndWriteResponse("Password is required", http.StatusBadRequest, w)
+			return
+		}
+		userAccount, err := env.DB.RetrieveUserByUsername(userRequest.Username)
+		if err != nil {
+			status := http.StatusInternalServerError
+			if errors.Is(err, certdb.ErrIdNotFound) {
+				status = http.StatusNotFound
+			}
+			logErrorAndWriteResponse(err.Error(), status, w)
+			return
+		}
+		if err := bcrypt.CompareHashAndPassword([]byte(userAccount.Password), []byte(userRequest.Password)); err != nil {
+			logErrorAndWriteResponse("Invalid password", http.StatusUnauthorized, w)
+			return
+		}
+		jwt, err := generateJWT(userRequest.Username, env.jwtSecret)
+		if err != nil {
+			logErrorAndWriteResponse(err.Error(), http.StatusInternalServerError, w)
+			return
+		}
+		response, err := json.Marshal(map[string]string{"jwt": jwt})
+		if err != nil {
+			logErrorAndWriteResponse("Failed to marshal JWT", http.StatusInternalServerError, w)
+			return
+		}
+		w.WriteHeader(http.StatusOK)
+		if _, err := w.Write(response); err != nil {
+			logErrorAndWriteResponse(err.Error(), http.StatusInternalServerError, w)
+		}
+	}
+}
+
 // logErrorAndWriteResponse is a helper function that logs any error and writes it back as an http response
 func logErrorAndWriteResponse(msg string, status int, w http.ResponseWriter) {
 	errMsg := fmt.Sprintf("error: %s", msg)
@@ -381,4 +432,18 @@ var GeneratePassword = func(length int) (string, error) {
 		b[i] = charset[n.Int64()]
 	}
 	return string(b), nil
+}
+
+// Helper function to generate a JWT
+func generateJWT(username, jwtSecret string) (string, error) {
+	token := jwt.NewWithClaims(jwt.SigningMethodHS256, jwt.MapClaims{
+		"username": username,
+		"exp":      time.Now().Add(time.Hour * 1).Unix(),
+	})
+	tokenString, err := token.SignedString([]byte(jwtSecret))
+	if err != nil {
+		return "", err
+	}
+
+	return tokenString, nil
 }
--- a/internal/api/handlers_test.go
+++ b/internal/api/handlers_test.go
@@ -101,10 +101,12 @@ const (
 )
 
 const (
-	adminUser      = `{"username": "testadmin", "password": "admin"}`
-	validUser      = `{"username": "testuser", "password": "user"}`
-	noPasswordUser = `{"username": "nopass", "password": ""}`
-	invalidUser    = `{"username": "", "password": ""}`
+	adminUser          = `{"username": "testadmin", "password": "admin"}`
+	validUser          = `{"username": "testuser", "password": "user"}`
+	invalidUser        = `{"username": "", "password": ""}`
+	noPasswordUser     = `{"username": "nopass", "password": ""}`
+	adminUserWrongPass = `{"username": "testadmin", "password": "wrongpass"}`
+	notExistingUser    = `{"username": "not_existing", "password": "user"}`
 )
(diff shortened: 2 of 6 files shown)


internal/api/handlers.go
```diff
@@ -360,6 +365,52 @@ func PostUserAccount(env *Environment) http.HandlerFunc {
...
+			if errors.Is(err, certdb.ErrIdNotFound) {
+				status = http.StatusNotFound
+			}
+			logErrorAndWriteResponse(err.Error(), status, w)
+			return
+		}
+		if err := bcrypt.CompareHashAndPassword([]byte(userAccount.Password), []byte(userRequest.Password)); err != nil {
+			logErrorAndWriteResponse("Invalid password", http.StatusUnauthorized, w)
+			return
+		}
```
kayra1, line 396: we should not specify if the error was with the password or the username for security reasons.

internal/api/handlers.go
```diff
@@ -382,3 +433,17 @@ var GeneratePassword = func(length int) (string, error) {
 	}
 	return string(b), nil
 }
+
+// Helper function to generate a JWT
+func generateJWT(username, jwtSecret string) (string, error) {
+	token := jwt.NewWithClaims(jwt.SigningMethodHS256, jwt.MapClaims{
```
kayra1, line 439: i think we mentioned using HMAC here.

internal/certdb/certdb.go
```diff
@@ -130,11 +131,11 @@ func (db *CertificateRequestsRepository) UpdateCSR(id string, cert string) (int6
 	if err != nil {
 		return 0, err
 	}
-	insertId, err := result.LastInsertId()
+	affectedRows, err := result.RowsAffected()
```
kayra1, line 134: since CSR's are unique, there will always be a single ID that was modified.

internal/certdb/certdb.go
```diff
@@ -219,11 +233,11 @@ func (db *CertificateRequestsRepository) UpdateUser(id, password string) (int64,
...
```
kayra1, line 236: same deal as above. that's why we only update passwords with id, to ensure uniqueness.

kayra1 (changes requested):
Great start

internal/certdb/certdb.go
saltiyazan: If there are 4 csrs in the db, and we modify the first one, the one with `ID=1` then `LastInsertId` would still return 5.
There is nothing wrong with that, but I don't see us using that information.

saltiyazan: But yes I agree affected rows will always return `1`

internal/api/handlers.go
saltiyazan: isn't `SigningMethodHS256` from hmac? or what do you mean?

internal/api/handlers.go
saltiyazan: True, fixed. Thanks

internal/api/handlers.go
kayra1: oh it is, nevermind

internal/certdb/certdb.go
kayra1: i thought it only returned that. then this is better.

internal/api/handlers.go
```diff
@@ -360,6 +365,53 @@ func PostUserAccount(env *Environment) http.HandlerFunc {
...
+		}
+		if userRequest.Password == "" {
+			logErrorAndWriteResponse("Password is required", http.StatusBadRequest, w)
+			return
+		}
+		userAccount, err := env.DB.RetrieveUserByUsername(userRequest.Username)
+		if err != nil {
+			status := http.StatusInternalServerError
+			if errors.Is(err, certdb.ErrIdNotFound) {
+				logErrorAndWriteResponse("The Username or Password is Incorrect. Try again.", http.StatusUnauthorized, w)
```
kayra1, line 388: small nit
```suggestion
				logErrorAndWriteResponse("The username or password is incorrect. Try again.", http.StatusUnauthorized, w)
```

internal/api/handlers.go
```diff
@@ -360,6 +365,53 @@ func PostUserAccount(env *Environment) http.HandlerFunc {
...
+				return
+			}
+			logErrorAndWriteResponse(err.Error(), status, w)
+			return
+		}
+		if err := bcrypt.CompareHashAndPassword([]byte(userAccount.Password), []byte(userRequest.Password)); err != nil {
+			logErrorAndWriteResponse("The Username or Password is Incorrect. Try again.", http.StatusUnauthorized, w)
```
kayra1, line 395:
```suggestion
			logErrorAndWriteResponse("The username or password is incorrect. Try again.", http.StatusUnauthorized, w)
```

internal/api/handlers.go
```diff
@@ -382,3 +434,17 @@ var GeneratePassword = func(length int) (string, error) {
 	}
 	return string(b), nil
 }
+
+// Helper function to generate a JWT
+func generateJWT(username, jwtSecret string) (string, error) {
+	token := jwt.NewWithClaims(jwt.SigningMethodHS256, jwt.MapClaims{
+		"username": username,
```
kayra1, line 441: lets put the permissions here too. this will help us avoid going to the database every time we need to authorize someone.

internal/certdb/certdb.go
saltiyazan: It will return the number of rows that have been modified, in this case `1`
I couldn't find a function that returns Ids. I'll take another look

internal/api/handlers.go
saltiyazan: ✅

internal/api/handlers.go
saltiyazan: ✅

internal/api/handlers.go
saltiyazan: ✅

kayra1 (changes requested):
almost done

internal/certdb/certdb.go
kayra1: I'm ok with changing this in a different PR

saltiyazan: cool

internal/certdb/certdb.go
```diff
@@ -30,11 +30,12 @@ const queryCreateUsersTable = `CREATE TABLE IF NOT EXISTS users (
...
 )`
 const (
-	queryGetAllUsers = "SELECT * FROM users"
-	queryGetUser     = "SELECT * FROM users WHERE user_id=?"
-	queryCreateUser  = "INSERT INTO users (username, password, permissions) VALUES (?, ?, ?)"
-	queryUpdateUser  = "UPDATE users SET password=? WHERE user_id=?"
-	queryDeleteUser  = "DELETE FROM users WHERE user_id=?"
+	queryGetAllUsers       = "SELECT * FROM users"
+	queryGetUser           = "SELECT * FROM users WHERE user_id=?"
+	queryGetUserByUsername = "SELECT * FROM users WHERE username=?"
```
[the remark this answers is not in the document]
saltiyazan: I think it's there.
This is it now `username TEXT NOT NULL UNIQUE,`

internal/api/handlers.go
```diff
@@ -360,6 +365,53 @@ func PostUserAccount(env *Environment) http.HandlerFunc {
...
+		if err := bcrypt.CompareHashAndPassword([]byte(userAccount.Password), []byte(userRequest.Password)); err != nil {
+			logErrorAndWriteResponse("The username or password is incorrect. Try again.", http.StatusUnauthorized, w)
+			return
+		}
+		jwt, err := generateJWT(userRequest.Username, env.jwtSecret, userAccount.Permissions)
+		if err != nil {
+			logErrorAndWriteResponse(err.Error(), http.StatusInternalServerError, w)
+			return
+		}
+		response, err := json.Marshal(map[string]string{"jwt": jwt})
```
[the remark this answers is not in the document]
saltiyazan: done

internal/api/handlers.go
```diff
@@ -360,6 +365,53 @@ func PostUserAccount(env *Environment) http.HandlerFunc {
 	}
 }
 
+func Login(env *Environment) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
```
kayra1, line 370: when sending just text, the content type should be "text/plain", or just not set.

internal/api/handlers.go
```diff
@@ -360,6 +365,53 @@ func PostUserAccount(env *Environment) http.HandlerFunc {
...
+		if err := bcrypt.CompareHashAndPassword([]byte(userAccount.Password), []byte(userRequest.Password)); err != nil {
+			logErrorAndWriteResponse("The username or password is incorrect. Try again.", http.StatusUnauthorized, w)
+			return
+		}
+		jwt, err := generateJWT(userRequest.Username, env.jwtSecret, userAccount.Permissions)
+		if err != nil {
+			logErrorAndWriteResponse(err.Error(), http.StatusInternalServerError, w)
+			return
+		}
+		response, err := json.Marshal(jwt)
```
kayra1, line 403: we don't have to marshal just jwt, we can just `w.Write(jwt)` below.

internal/api/handlers.go
saltiyazan: ✅

internal/api/handlers.go
saltiyazan: ✅

internal/api/handlers.go
```diff
@@ -404,3 +450,18 @@ var GeneratePassword = func(length int) (string, error) {
 	}
 	return string(b), nil
 }
+
+// Helper function to generate a JWT
+var GenerateJWT = func(username, jwtSecret string, permissions int) (string, error) {
```
kayra1, line 455: lets not do function declarations like this unless its anonymous or higher order
```suggestion
function GenerateJWT(username, jwtSecret string, permissions int) (string, error) {
```

kayra1 (approved):
one last nit

internal/api/handlers_test.go
```diff
@@ -375,6 +377,12 @@ func TestGoCertUsersHandlers(t *testing.T) {
 	}
 	defer func() { server.GeneratePassword = originalFunc }()
 
+	originalJwtFunc := server.GenerateJWT
+	server.GenerateJWT = func(username, jwtSecret string, permissions int) (string, error) {
+		return "jwt", nil
+	}
+	defer func() { server.GenerateJWT = originalJwtFunc }()
```
kayra1, line 384: we decided to avoid introducing mocks

kayra1 (changes requested):

internal/api/handlers.go

internal/api/handlers_test.go
saltiyazan: Done

kayra1 (approved):
great job

PR #42, opened 2024-07-10
feat: authentication middleware for the GoCert server

kayra1:
# Description

This PR implements the authentication and permission functionality described in TE091. The `/metrics` and `/status` endpoints are left unauthenticated, as well as the first POST request to the `api/v1/accounts` endpoint. For more details, please read TE091.

# Checklist:

### DIFF
--- a/internal/api/handlers_test.go
+++ b/internal/api/handlers_test.go
@@ -127,6 +127,10 @@ func TestGoCertCertificatesHandlers(t *testing.T) {
 
 	client := ts.Client()
 
+	var adminToken string
+	var nonAdminToken string
+	t.Run("prepare user accounts and tokens", prepareUserAccounts(ts.URL, client, &adminToken, &nonAdminToken))
+
 	testCases := []struct {
 		desc     string
 		method   string
@@ -347,6 +351,7 @@ func TestGoCertCertificatesHandlers(t *testing.T) {
 	for _, tC := range testCases {
 		t.Run(tC.desc, func(t *testing.T) {
 			req, err := http.NewRequest(tC.method, ts.URL+tC.path, strings.NewReader(tC.data))
+			req.Header.Set("Authorization", "Bearer "+adminToken)
 			if err != nil {
 				t.Fatal(err)
 			}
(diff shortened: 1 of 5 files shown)


internal/api/middleware.go
```diff
@@ -84,3 +89,76 @@ func loggingMiddleware(ctx *middlewareContext) middleware {
...
+		{"POST", `accounts\/\d+\/change_password$`},
+	}
+
+	return func(next http.Handler) http.Handler {
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			if !strings.HasPrefix(r.URL.Path, "/api/v1/") {
+				next.ServeHTTP(w, r)
+				return
+			}
+			if r.Method == "POST" && strings.HasSuffix(r.URL.Path, "accounts") && !ctx.firstAccountIssued {
```
saltiyazan, line 110: Nitpicking but does the `r.Method == "POST"` part have any value here?
It might be a nicer experience to get an empty list of user when sending a GET to /accounts when no users are created yet.

internal/api/middleware.go
```diff
@@ -84,3 +89,76 @@ func loggingMiddleware(ctx *middlewareContext) middleware {
...
+				ctx.firstAccountIssued = true
```
saltiyazan, line 111: Providing an invalid password (for example too short) on the first attempt to create the admin user will result in this being true without a user created and there is no way to recover.

saltiyazan (changes requested):
Looks good, I ran into one issue while testing that I commented

internal/api/middleware.go
kayra1: just to be clear are you saying users who haven't logged in yet should be able to get the list of accounts if no account has been created yet?

internal/api/middleware.go
kayra1: great catch, fixed.

internal/api/middleware.go
saltiyazan: Yes, only if no accounts have been created
But it doesn't make any difference, as the result will be no accounts in both cases
I only think it could be a nicer experience
But as I think of it a bit more I realise it doesn't bring any added value, so ignore it if you want :)

saltiyazan (approved):

internal/api/handlers_test.go
```diff
@@ -379,43 +384,43 @@ func TestGoCertUsersHandlers(t *testing.T) {
...
 		status   int
 	}{
-		{
-			desc:     "Create first user success",
-			method:   "POST",
-			path:     "/api/v1/accounts",
-			data:     adminUser,
-			response: "{\"id\":1}",
-			status:   http.StatusCreated,
-		},
```
saltiyazan, line 397: Why are we removing this test case?

kayra1: it is included in the previous test case that creates 2 accounts, logs in as them and saves their login tokens for the rest of the test suite

internal/api/middleware.go
kayra1: I think it would be a bit weird to have that send an empty response instead of a 403 for the entire `/api/v1`. The POST request is needed but this one is not so it may give the wrong impression.

PR #43, opened 2024-07-12
feat: redirect /accounts/me to the correct /accounts/{id}

kayra1:
# Description

This change introduces a small change to the endpoints where users are allowed to use `/me` to refer to their own accounts. It also cleans up some of the auth middleware code to be a bit reusable.

# Checklist:


internal/api/handlers.go
```diff
@@ -489,6 +510,7 @@ func logErrorAndWriteResponse(msg string, status int, w http.ResponseWriter) {
 	log.Println(errMsg)
 	w.WriteHeader(status)
 	if _, err := w.Write([]byte(errMsg)); err != nil {
+		// TODO: how did this get merged?
```
ghislainbourgeois, line 513: Should we fix this right away? It seems like a potential infinite loop that could crash the server.

Otherwise, is there a ticket to track this?

internal/api/middleware.go
```diff
@@ -152,6 +148,45 @@ func authMiddleware(ctx *middlewareContext) middleware {
...
+		return nil, fmt.Errorf("authorization header couldn't be processed. The expected format is 'Bearer <token>'")
+	}
+	claims, err := getClaimsFromJWT(bearerToken[1], jwtSecret)
+	if err != nil {
+		return nil, fmt.Errorf("token is not valid: %s", err)
+	}
+	return claims, nil
+}
+
+func permitter(claims *jwtGocertClaims, regex, path string, SelfAuthorizedAllowed bool) (pathMatched, idMatchedIfExistsInPath bool, err error) {
```
ghislainbourgeois, line 166: This whole function is a bit unclear in what it is supposed to do at a glance. The name is quite generic, and it seems to return booleans that things matched in the accessed path, and not permit or deny anything.

I think the name could be improved and a docstring would be helpful here.

ghislainbourgeois (dismissed):

internal/api/handlers.go
kayra1: fixed. lets not leave this here haha.

internal/api/middleware.go
kayra1: reworked it a bit more. To be honest this entire thing could be improved a lot but it's difficult to avoid writing a new go web framework without having the code be ugly and difficult to read.

internal/api/middleware.go
```diff
@@ -152,6 +144,63 @@ func authMiddleware(ctx *middlewareContext) middleware {
...
+//
+// There are two types of restricted paths: admin only paths that only admins can access, and self authorized paths,
+// which users are allowed to use only if they are taking an action on their own user ID. The second question answers the
+// question "If the path requires an ID, is the user attempting to access their own ID?"
+//
+// For all endpoints and permission permutations, there are only 2 cases when users are allowed to use endpoints:
+// If the URL path is not restricted to admins
+// If the URL path is restricted to self authorized endpoints, and the user is taking action to their own ID
+// This function returns if the URL matches the restricted path, and if that request should be allowed according to the conditions above.
+func AllowRequest(claims *jwtGocertClaims, regex, path string, SelfAuthorizedAllowed bool) (pathMatched, requestAllowed bool, err error) {
```
gruyaume, line 173: Why used named return values here (and not anywhere else)? I'd recommend to define them in the func, for consistency purposes

internal/api/middleware.go
```diff
@@ -115,35 +117,25 @@ func authMiddleware(ctx *middlewareContext) middleware {
...
 					}
-					if r.Method == v.method && matched {
+					if pathMatched && !requestAllowed {
 						logErrorAndWriteResponse("forbidden", http.StatusForbidden, w)
 						return
 					}
 				}
 			}
-			if claims.Permissions == 1 && r.Method == "DELETE" && strings.HasSuffix(r.URL.Path, "accounts/1") {
+			if r.Method == "DELETE" && strings.HasSuffix(r.URL.Path, "accounts/1") {
```
gruyaume, line 138: Why does this require a special case? Could it not use a regex like the other restricted paths?

internal/api/middleware.go
gruyaume, line 173: Do we really need to return this `pathMatched`? Could it not simply be forbidden if the path does not match?

gruyaume (changes requested):

internal/api/middleware.go
kayra1: changed

internal/api/middleware.go
kayra1: This is special because even the admin user is restricted to this path.

internal/api/middleware.go
kayra1: I reworked this function again a little bit. It's starting to make sense more. What do you think?

internal/api/middleware.go
```diff
@@ -152,6 +147,67 @@ func authMiddleware(ctx *middlewareContext) middleware {
...
+//
+// There are two types of restricted paths: admin only paths that only admins can access, and self authorized paths,
+// which users are allowed to use only if they are taking an action on their own user ID. The second question is
+// "If the path requires an ID, is the user attempting to access their own ID?"
+//
+// For all endpoints and permission permutations, there are only 2 cases when users are allowed to use endpoints:
+// If the URL path is not restricted to admins
+// If the URL path is restricted to self authorized endpoints, and the user is taking action with their own ID
+// This function validates that the user the with the given claims is allowed to use the endpoints by passing the above checks.
+func AllowRequest(claims *jwtGocertClaims, method, path string, permissionRestrictions []struct {
```
gruyaume, line 176: RestrictedPaths is only used inside of this func right? Why not have it inside of the func instead of taking it as a parameter. Then the inputs can only be claims, path and the returns can only be bool, error.

gruyaume (changes requested):

internal/api/middleware.go
kayra1: yep. done

gruyaume (approved):

PR #46, opened 2024-07-24
feat: change password form

kayra1:
# Description

This change lets users change their account password

image

# Checklist:

### DIFF
--- a/ui/src/app/change_password/page.tsx
+++ b/ui/src/app/change_password/page.tsx
@@ -48,7 +48,7 @@ export default function ChangePasswordPage() {
                 <div className="p-panel__content">
                     <div className="u-fixed-width">
                         <form className={"p-form-validation " + (!passwordsMatch ? "is-error" : "")}>
-                            <label className="p-form__label">Password</label>
+                            <label className="p-form__label">New Password</label>
                             <input className="p-form-validation__input" type="password" id="password1" name="password" placeholder="******" autoComplete="current-password" required={true} onChange={handlePassword1Change} />
                             <p className="p-form-help-text">
                                 Password must have 8 or more characters, must include at least one capital letter, one lowercase letter, and either a number or a symbol.
--- a/ui/src/app/change_password/page.tsx
+++ b/ui/src/app/change_password/page.tsx
@@ -46,7 +46,7 @@ export default function ChangePasswordPage() {
             }}>
                 <div className="p-panel__content">
                     <div className="u-fixed-width">
-                        <form className={"p-form-validation " + (!passwordsMatch ? "is-error" : "")}>
+                        <form className={"p-form-validation " + ((!passwordsMatch || !passwordIsValid(password1)) && password2 != "" ? "is-error" : "")}>
                             <fieldset>
                                 <h4 className="p-panel__title">Change Password</h4>
                                 <label className="p-form__label">New Password</label>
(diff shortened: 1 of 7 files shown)


ui/src/app/change_password/page.tsx
```diff
@@ -0,0 +1,81 @@
...
+                                        <p className="p-notification__message">{errorText.split("error: ")}</p>
+                                    </div>
+                                </div>
+                            }
+                            {!passwordsMatch ? (
+                                <>
+                                    <button type="submit" name="submit" disabled={true}>Submit</button>
+                                </>
+                            ) : (
+                                <button type="submit" name="submit" onClick={(event) => { event.preventDefault(); mutation.mutate({ authToken: (auth.user ? auth.user.authToken : ""), username: (auth.user ? auth.user.username : ""), password: password1 }) }}>Submit</button>
```
gruyaume, line 72: Can we move the submit button to the right and make it green?

I listed enhancement recommendations for the login page and some of those can be applied here as well. Including the submit button position and color. 

- https://github.com/canonical/gocert/issues/47

ui/src/app/change_password/page.tsx
```diff
@@ -0,0 +1,81 @@
...
+                minWidth: "min-content",
+                minHeight: "min-content",
+            }}>
+                <div className="p-panel__header is-sticky">
+                    <h4 className="p-panel__title">Change Password</h4>
+                </div>
+                <div className="p-panel__content">
+                    <div className="u-fixed-width">
+                        <form className={"p-form-validation " + (!passwordsMatch ? "is-error" : "")}>
+                            <label className="p-form__label">Password</label>
```
gruyaume, line 51:
```suggestion
                            <label className="p-form__label">New Password</label>
```

ui/src/app/change_password/page.tsx
```diff
@@ -0,0 +1,81 @@
...
+                            <input className="p-form-validation__input" type="password" id="password1" name="password" placeholder="******" autoComplete="current-password" required={true} onChange={handlePassword1Change} />
+                            <p className="p-form-help-text">
+                                Password must have 8 or more characters, must include at least one capital letter, one lowercase letter, and either a number or a symbol.
+                            </p>
+                            <label htmlFor="p-form__label">Password Again</label>
```
gruyaume, line 56:
```suggestion
                            <label htmlFor="p-form__label">Confirm New Password</label>
```

gruyaume (changes requested):

ui/src/app/change_password/page.tsx
kayra1: I've added some of the changes from there to this form, please take a look again.

The only thing from this issue i may push back on is putting the button on the right. English text starts from the left, so it's usual for readers to look at the left of the row for the button under a form. I personally like it there too.

also added a show button to see the password for the first field.

I will take this issue's ticket after all of the queued up changes are merged.

gruyaume: I'm ok with keeping it on the left.

ui/src/app/change_password/page.tsx
```diff
@@ -0,0 +1,100 @@
...
+                width: "45rem",
+                minWidth: "min-content",
+                minHeight: "min-content",
+            }}>
+                <div className="p-panel__content">
+                    <div className="u-fixed-width">
+                        <form className={"p-form-validation " + (!passwordsMatch ? "is-error" : "")}>
+                            <fieldset>
+                                <h4 className="p-panel__title">Change Password</h4>
+                                <label className="p-form__label">New Password</label>
```
gruyaume, line 52: I'm not a huge fan of the fact that as soon as you start working the form becomes red. I haven't done anything wrong yet.

gruyaume, line 52: When the password does not have the requirements (ex. capital letter) it's not possible to submit but there is no red badge to tell us why we did something wrong.

gruyaume (changes requested):

ui/src/app/change_password/page.tsx
kayra1: I implemented 
> Have the button disabled until the input values become acceptable
from the issue. But if i do this and the above comment, then the form will become red immediately.

A good middle ground could be to only have an error once the user starts inputting text to the second field. I've made the changes, please take a look.

kayra1: I could make it so the errors with the password validation show up first. It would behave very similarly to updating when unfocused.

There isn't really a standard for this, I can find examples for every version. It's just about what we want.

kayra1: I've made some changes, please check again.

gruyaume: This feels better in my opinion

gruyaume (approved):

PR #48, opened 2024-07-25
feat: users table

kayra1:
# Description

This change adds the user table and the accompanying aside form to add a new user to GoCert. The tab is only visible to an Admin user, and a regular user will only see the certificate requests table.

image

image

image

image

# Checklist:


DanielArndt (approved):
This is the first typescript I've looked at in years, so take my approval with a grain of salt...

ui/src/app/users/asideForm.tsx
```diff
@@ -0,0 +1,105 @@
...
+            <div className="p-panel__header">
+                <h4 className="p-panel__title">Add a New User</h4>
+                <div className="p-panel__controls">
+                    <button onClick={() => asideContext.setIsOpen(false)} className="p-button--base u-no-margin--bottom has-icon"><i className="p-icon--close"></i></button>
+                </div>
+            </div>
+            <div className="p-panel__content">
+                <form className={"p-form-validation " + ((!passwordIsValid(password1) && password1 != "") || (!passwordsMatch && password2 != "") ? "is-error" : "")}>
+                    <div className="p-form__group row">
+                        <label className="p-form__label">User Name</label>
```
gruyaume, line 42:
```suggestion
                        <label className="p-form__label">Username</label>
```

ui/src/app/users/asideForm.tsx
```diff
@@ -0,0 +1,105 @@
...
+                            <div className="p-notification--negative">
+                                <div className="p-notification__content">
+                                    <h5 className="p-notification__title">Error</h5>
+                                    <p className="p-notification__message">{errorText.split("error: ")}</p>
+                                </div>
+                            </div>
+                        }
+                        {!passwordsMatch || !passwordIsValid(password1) ? (
+                            <>
+                                <button type="submit" name="submit" disabled={true}>Submit</button>
```
gruyaume, line 83: Can we make the submit button green (for consistency)?

ui/src/app/aside.tsx
gruyaume, line 1: I'm getting this error when trying to add a new user:

image

## Go logs

```
2024/07/30 16:18:28 error: auth failed: authorization header not found
```

ui/src/app/users/table.tsx
```diff
@@ -0,0 +1,42 @@
...
+                </div>
+            </div>
+            <div className="p-panel__content">
+                <div className="u-fixed-width">
+                    <table id="csr-table" aria-label="Certificate Requests Table" className="p-table--expanding">
+                        <thead>
+                            <tr>
+                                <th>ID</th>
+                                <th>Username</th>
+                                <th>Delete</th>
```
gruyaume, line 27: Can we take the same approach we have for CSR's and have a clickable menu icon which would lead to the Delete option. We'll likely want the Change Password option in there as well.

gruyaume (changes requested):

ui/src/app/users/table.tsx
kayra1: Do we want the admin to be able to change the password for any user?

ui/src/app/aside.tsx
kayra1: This is weird, it looks like i didn't add the auth header to the query, but it was somehow succeeding locally. Should be fixed.

ui/src/app/users/asideForm.tsx
kayra1: done

ui/src/app/users/asideForm.tsx
```diff
@@ -0,0 +1,105 @@
...
+// function SubmitUser({ csrText, onClickFunc }: { csrText: string, onClickFunc: any }) {
+//     const validationComponent = csrText == "" ? <></> : csrIsValid ? <div><i className="p-icon--success"></i>Valid CSR</div> : <div><i className="p-icon--error"></i>Invalid CSR</div>
+//     const buttonComponent = csrIsValid ? <button className="p-button--positive u-float-right" name="submit" onClick={onClickFunc} >Submit</button> : <button className="p-button--positive u-float-right" name="submit" disabled={true} onClick={onClickFunc} >Submit</button>
+//     return (
+//         <>
+//             {validationComponent}
+//             {buttonComponent}
+//         </>
+//     )
+// }
```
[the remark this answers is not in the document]
kayra1: oops. gone

ui/src/app/users/table.tsx
kayra1: made the actions menu anyway since just having a delete button looked awkward

ui/src/app/users/components.tsx
```diff
@@ -0,0 +1,32 @@
...
+    onMouseDownFunc: () => void
+    warningText: string
+} | null
+
+interface ConfirmationModalProps {
+    modalData: ConfirmationModalData
+    setModalData: Dispatch<SetStateAction<ConfirmationModalData>>
+}
+
+export function ConfirmationModal({ modalData, setModalData }: ConfirmationModalProps) {
```
gruyaume, line 13: We should probably specify what is the user that is being deleted in the configmation modal

ui/src/app/users/table.tsx
gruyaume: That's a good question (Related to changing passwords). I'm not sure what the bet practice is relating to this.

ui/src/app/aside.tsx
gruyaume: ok it works now

ui/src/app/users/table.tsx
kayra1: done. I've added this change directly to this PR, please check it out.

ui/src/app/users/components.tsx
kayra1: updated

gruyaume (changes requested):

ui/src/app/nav.tsx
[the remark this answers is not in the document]
kayra1: non-admin users can't see the users page, so for them the aside would show up on the right side of the certificate table, which would be confusing imo. We shouldn't have the aside be the spot for every form, it should be for the main action on the page it is on. How about a modal popup instead, for both regular users and admin users?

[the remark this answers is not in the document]
gruyaume: I'm ok if it's a modal instead of an aside, I just want to avoid having two different views for doing the same action.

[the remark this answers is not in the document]
kayra1: Updated. please check again.

gruyaume (approved):

PR #60, opened 2024-08-22
feat!: accept bundles instead of just certificates

kayra1:
# Description

This PR adds the change required to accept and distribute issuer certificates alongside the certificate of the CSR. A new certificate must contain a bundle of at least 2 certificates in a valid chain to be accepted. Any extra characters will be ignored.

# Checklist:

### DIFF
--- a/internal/api/handlers_test.go
+++ b/internal/api/handlers_test.go
@@ -16,91 +16,128 @@ import (
 )
 
 const (
-	validCSR1 = `-----BEGIN CERTIFICATE REQUEST-----
-MIICszCCAZsCAQAwFjEUMBIGA1UEAwwLZXhhbXBsZS5jb20wggEiMA0GCSqGSIb3
-DQEBAQUAA4IBDwAwggEKAoIBAQDC5KgrADpuOUPwSh0YLmpWF66VTcciIGC2HcGn
-oJknL7pm5q9qhfWGIdvKKlIA6cBB32jPd0QcYDsx7+AvzEvBuO7mq7v2Q1sPU4Q+
-L0s2pLJges6/cnDWvk/p5eBjDLOqHhUNzpMUga9SgIod8yymTZm3eqQvt1ABdwTg
-FzBs5QdSm2Ny1fEbbcRE+Rv5rqXyJb2isXSujzSuS22VqslDIyqnY5WaLg+pjZyR
-+0j13ecJsdh6/MJMUZWheimV2Yv7SFtxzFwbzBMO9YFS098sy4F896eBHLNe9cUC
-+d1JDtLaewlMogjHBHAxmP54dhe6vvc78anElKKP4hm5N5nlAgMBAAGgWDBWBgkq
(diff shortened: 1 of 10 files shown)


internal/certdb/validation.go
```diff
@@ -27,27 +30,58 @@ func ValidateCertificateRequest(csr string) error {
...
-		return errors.New("given PEM string not a certificate")
+	if len(certificates) < 2 {
+		return errors.New("less than 2 certificate PEM strings were found")
 	}
-	_, err := x509.ParseCertificate(certBlock.Bytes)
-	if err != nil {
-		return err
+	for i := 0; i < len(certificates)-1; i++ {
+		firstCert := certificates[i]
+		secondCert := certificates[i+1]
```
ghislainbourgeois, line 64:
```suggestion
	for i, firstCert := range certificates[:len(certificates)-1] {
		secondCert := certificates[i+1]
```

internal/certdb/validation.go
```diff
@@ -27,27 +30,58 @@ func ValidateCertificateRequest(csr string) error {
...
+		if !bytes.Equal(firstCert.RawIssuer, secondCert.RawSubject) {
+			return fmt.Errorf("invalid certificate chain: certificate %d, certificate %d: subjects do not match", i, i+1)
+		}
+		if err := firstCert.CheckSignatureFrom(secondCert); err != nil {
+			return fmt.Errorf("invalid certificate chain: certificate %d, certificate %d: keys do not match: %s", i, i+1, err.Error())
+		}
+	}
+	for i := 1; i < len(certificates)-1; i++ {
+		cert := certificates[i]
```
ghislainbourgeois, line 73:
```suggestion
	for i, cert := range certificates[:len(certificates)-1] {
```

internal/api/handlers_test.go
```diff
@@ -349,7 +413,7 @@ func TestGoCertCertificatesHandlers(t *testing.T) {
 		},
 	}
 	for _, tC := range testCases {
-		t.Run(tC.desc, func(t *testing.T) {
+		t.Run(fmt.Sprintf("step %d: %s", tC.step, tC.desc), func(t *testing.T) {
```
ghislainbourgeois, line 416: I would remove the step field and simply take this from the range:

```suggestion
    for step, tC := range testCases {
		t.Run(fmt.Sprintf("step %d: %s", tC.step, tC.desc), func(t *testing.T) {
```

ghislainbourgeois (changes requested):

internal/api/handlers_test.go
kayra1: I liked having the steps because it helped scrolling to find the exact case that went wrong.

ghislainbourgeois: Can't the same be done achieved with the description? This is making the file 26 lines longer at the moment, and will be a pain to maintain cleanly if we decide to remove a test case.

kayra1: that's not a bad idea. changed.

internal/certdb/validation.go
kayra1: this one actually starts at 1, so i used a range loop with the slice starting from 1.

ghislainbourgeois (approved):

PR #74, opened 2024-09-09
feat: allow for multiple cli commands

gruyaume:
# Description

We add the option to have multiple commands in notary. For now we only implement the `start` command with the same usage as the initial root command:

```shell
notary start -config config.yaml
``` 

## Rationale

We will soon add a `version` command which should be handled differently from starting the server and will not require a config file for instance. Here we will want to run:

```shell
notary version
notary start --config config.yaml
```

And those two commands should be handled independently.

## This is a breaking change

This is a breaking change, and will require changes in the charm. Here we only bump the minor version as Notary is not yet released and we don't want the first released version to be 32.51.1.

# Checklist:


kayra1:
What about adding a version field to the `/status` API endpoint? It's a bit more cumbersome to use, but I just can't imagine us having any other commands in the CLI. notary is a web application after all, and this is its binary.

gruyaume:
> What about adding a version field to the `/status` API endpoint? It's a bit more cumbersome to use, but I just can't imagine us having any other commands in the CLI. notary is a web application after all, and this is its binary.

If someone is running notary on their system and quickly want to know which version they are running, I don't think we want to ask them to make an API call. Though I agree it could be surfaced in the UI somewhere.

I can definitely imagine it having a cli to communicate with the notary server. I'm not saying we will, but I can see some value in something like:

```
notary login <my token>
notary request certificate <my csr>
notary version
notary help
```

internal/cli/start.go
```diff
@@ -0,0 +1,52 @@
...
+	return startCommand.fs.Name()
+}
+
+func (startCommand *StartCommand) Init(args []string) error {
+	return startCommand.fs.Parse(args)
+}
+
+func (startCommand *StartCommand) Run() error {
+	if startCommand.configPath == "" {
+		return fmt.Errorf("providing a config file is required")
```
saltiyazan, line 27: I was reading and apparently it's better to use `%w` for errors.
It wraps the error when returning allowing the caller to unwrap it.
https://go.dev/wiki/ErrorValueFAQ

So basically the caller of Run would be able to check which error was returned out of the 4 here.
We probably won't use it because we just are going to exit and log the error whichever that was, therefore, it's fine not to change it, just thought it was interesting to share.

saltiyazan (approved):
Looks good.
Feel free to resolve comments and merge.

kayra1:
> I can definitely imagine it having a cli to communicate with the notary server. I'm not saying we will, but I can see some value in something like:
> 
> ```
> notary login <my token>
> notary request certificate <my csr>
> notary version
> notary help
> ```

that may be nice actually

internal/cli/start.go
gruyaume: Done

internal/cli/cli.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+	return allowedSubcommands
+}
+
+func Run(args []string) error {
+	if len(args) < 1 {
+		return fmt.Errorf("you must pass a subcommand. Allowed commands: %v", getSupportedCommandNames())
+	}
+
+	subcommand := os.Args[1]
```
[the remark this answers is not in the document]
gruyaume: No good reason, I moved to using the func args

gruyaume:
Moving to draft as I am improving this

PR #93, opened 2024-09-24
feat: log to stderr and graceful shutdown

kayra1:
# Description

This PR reverts the change to log to stdout and listens to SIGINT to shutdown the server gracefully by waiting for the termination of all connection

# Checklist:


internal/config/config.go
```diff
@@ -25,49 +27,55 @@ type Config struct {
 
 // Validate opens and processes the given yaml file, and catches errors in the process
 func Validate(filePath string) (Config, error) {
-	validationErr := errors.New("config file validation failed: ")
+	const validationErr = "config file validation failed: %w"
```
gruyaume, line 30: Let's not prefix errors, it's the job of the parent

internal/config/config.go
```diff
@@ -25,49 +27,55 @@ type Config struct {
...
 	config := Config{}
 	configYaml, err := os.ReadFile(filePath)
 	if err != nil {
-		return config, errors.Join(validationErr, err)
+		return Config{}, fmt.Errorf(validationErr, err)
```
gruyaume, line 34:
```suggestion
		return Config{}, "cannot read config file")
```

gruyaume (changes requested):
a couple of minor improvements, looks good otherwise

internal/config/config.go
kayra1: done

internal/config/config.go
kayra1: not done

gruyaume (approved):

Issue #66, opened 2024-09-06
Add a `--version` to notary CLI


gruyaume:
### Enhancement Proposal

As a notary user I would like to quickly be aware of which version of the software I am using. This feature will be helpful in answering the following questions:
- Am I affected by bug xyz
- Am I running the latest version

It will also be useful in setting up the workload version in the notary k8s charm.

gruyaume:
The first part of doing this is to add the concept of cli commands in Notary

gruyaume:
I removed the good first issue tag as this is not as trivial as I anticipated and required an important project structure change. Reference:
- #74

PR #95, opened 2024-09-24
feat: add version to /status endpoint

gruyaume:
# Description

We return the Notary version part of the return response made to `/status`. We also leverage snapcraft's `adopt-info` feature to set the snap and rock versions from that file. This avoid having to set the project's version in multiple places.

This is the first step in addressing #66 , where the next step is the frontend effort to view this version.

### Users know which version they are running

This endpoint will be used by the frontend (in follow-up PR's) so that users can quickly know which versions of Notary they are running. This is important for users to be able to know if:
- They are running the latest Notary version
- They are affected by a bug

### Centralized version management
Notary's version is now centrally managed in `version/VERSION`, similarly to other mainstream Go projects (like Vault). The snap and rock simply read this file to figure their own version out.

### Easier releasing

Very soon we'll want to leverage Github releases and this will be easy to do with this VERSION file, all we'll have to do is add a `-dev` suffix and when a PR removes the suffix (ex. `0.0.1-dev` -> `0.0.1`), the CI will automatically:
- Create the GitHub release
- Create a PR that bumps the version and re-adds the suffix (ex. `0.0.1` -> `0.0.2-dev`). 

## Reference
- https://snapcraft.io/docs/using-external-metadata

# Checklist:


kayra1 (commented):
yay :D

gruyaume:
> yay :D

Is this an approval ? :laughing:

kayra1 (approved):
oops, it should have been :D

PR #113, opened 2024-10-18
chore: use route groups to organize endpoints

kayra1:
# Description

This change does a couple things, mainly moving things around:
* Uses route groups with separate layouts, one with and one without navigation. This allows us to remove the check in navigation for '/login' and '/initialize'.
* Uses the @ sign to indicate project root for imports so we don't have the `../../../../../` chain in the imports.
* Moves some common components outside of the app folder into a components folder
* Moves non `.tsx` files like `types.ts` and `queries.ts` files outside of the UI folder, since they are not UI files.
* Moves some very common files like css files to the top level `src` folder.

# Checklist:


gruyaume (approved):
Looks good, I haven't tested it locally, I assume you have and everything is still in place!

PR #122, opened 2024-11-12
feat: replace pure SQL strings with SQLair

### LINKED ISSUE #84: Ben's style and structure review
### Enhancement Proposal

Overall, this is nicely structured! Good package structure, excellent use of the standard library rather than tons of 3rd party depenendencies, some good tests, and so on. If this is your first Go project, well done!

I've written up a bunch of comments below, labelling each "Minor", "Major", or "Catastrophic" (none yet -- might have some next week :-). The difference between Major and Minor is a bit arbitrary, but I've mostly used Major for "strong recommendation" and Minor for "style stuff". That said, many of them would be quick refactorings or quick fixes, so I'd still recommend doing that to tidy up the style a bit.

A couple of the more interesting things:

* "only apply middleware to the handlers that need them", rather than applying them to all routes and then doing ad-hoc filtering and route matching in the middlewares.
* I haven't reviewed `db.go` yet. But instead of doing your own database row scanning and parametrisation, consider using sqlc (I've used it in personal projects and really recommmend it) or Canonical's own [sqlair](https://github.com/canonical/sqlair) type mapper. I like sqlc because it's pre-generated from your db schema with `go generate` so you get nice type-safe structs. I can give you a demo of sqlc if you want.

I haven't finished the review. I have some security concerns about how you're generating passwords -- but that's enough for today! I'll review `handlers_users.go` and the database code next week.

Early notes from a quick scan of the database code: `SELECT *` is a bad idea in production code -- I'll explain why next week. I've also got some comments/concerns about the database schema.

Overall, it'd definitely be worthwhile reading a couple of Go style guides in detail, obviously the Canonical one (from Ed Jones), and also [Google's Go style guide](https://google.github.io/styleguide/go/) is good.

## cmd/notary/main.go

**Minor:** Why do you want the logs going to stdout? It's more typical to have logs going to stderr.

```go
	log.SetOutput(os.Stdout)
```

**Minor:** You might want to look into graceful shutdown, so that if SIGTERM is sent (or Ctrl-C is pressed), it allows a small amount of time for requests to finish before exiting. This is pretty easy to achieve in Go. There's an example of this pattern in the [`Server.Shutdown`](https://pkg.go.dev/net/http#Server.Shutdown) docs.

## cmd/notary/main_test.go

**Minor:** Would be nice if TestMain didn't install stuff into your go home directory. When we've done this, we've just done a `go build ./cmd/notary` to build the binary in the current directory, and then use that.

```go
func TestMain(m *testing.M) {
	cmd := exec.Command("go", "install", "./...")
```

**Minor:** It's better to follow the standard `err` naming (rather than `writeCertErr` or `writeConfigErr`) and check the error after each call, even if a tiny bit more verbose:

```go
	writeCertErr := os.WriteFile(testfolder+"/cert_test.pem", []byte(validCert), 0o644)
	writeKeyErr := os.WriteFile(testfolder+"/key_test.pem", []byte(validPK), 0o644)
	if writeCertErr != nil || writeKeyErr != nil {
		log.Fatalf("couldn't create temp testing file")
	}
```

That said, it might be better to create a `cmd/notary/testdata` directory and have those cert and key files in there -- it makes the test file a bit cleaner, you won't need to write them out, and `go test` automatically changes the current directory to the package directory when running the test binary, so you can reference them with a relative path like `testdata/cert.pem`. You can probably get rid of all the temp directory and cleanup stuff in `TestMain` in that case too.

**Major:** For all of these errors, you should include the `err` value in the log message, so that you're not completely blind if something does go wrong. Similar throughout. For example, change the above to:

```go
		log.Fatalf("couldn't create temp testing file: %v", err)
```

**Minor:** Nice use of table-driven tests. However, you should use [`t.Run`](https://pkg.go.dev/testing#T.Run) inside the test cases loop so that you can observe and select specific sub-tests with `-run=Foo/bar. I see you've done that in some other places, just not here.

**Minor:** I don't think you need to save/restore `os.Args` here, as you're running a new command, not doing anything with `os.Args`. You can remove this:

```go
	oldArgs := os.Args
	defer func() { os.Args = oldArgs }()
```

**Minor:** In `TestNotaryFail`, you can probably use [`Cmd.CombinedOutput`](https://pkg.go.dev/os/exec#Cmd.CombinedOutput) to avoid the Start/Wait/read-StdoutPipe and simplify the code.

## internal/api/server.go

**Major:** It's highly unusual (and confusing I think) to have a directory named something different from the Go package name (`internal/api` is the directory, but the package is `server`). Leading to you wanting to rename the import:

```
	server "github.com/canonical/notary/internal/api"
```

Just rename the directory to `server` and avoid the confusion.

**Minor:** In Go everything is prefixed with the package name, so you want to make use of that and avoid "stuttering" if you can. So if there's one main type, use `server.New()` rather than `server.NewServer()`, and `db.New()` rather than `db.NewDatabase()`.

For `server.NewNotaryRouter`, that should probably be `server.NewRouter` (it's redundant to include the project name).

**Minor:** `Environment` seems like a confusing name (it's not environment variables). How about `RouterConfig` or `HandlerConfig`? (And rename `NewRouter` to `NewHandler` -- which is actually more what it is, as it does all the handling, not just routing.)

**Minor:** In `NewServer`, why does creating the key pair return an error (good) but connecting to the database does a `log.Fatalf` (not so good)?

**Minor:** In `SendPebbleNotification`, this should be just a wrapped error, no need for errors/New+errors.Join:

```go
		return errors.Join(errors.New("couldn't execute a pebble notify: "), err)
```

Should be just `fmt.Errorf` with `%w`:

```go
		return fmt.Errorf("couldn't execute pebble notify: %w", err)
```

I see you've done that elsewhere though.

## internal/api/server_test.go

**Minor:** It's unusual to use `TestMain` in this way. Is it actually used for anything in this package? It looks like you're not actually using these files, and you could probably remove `TestMain` altogether (it's usually not needed at all).

If anything, it's much more common to have explicit setup helper functions, using `t.Helper()` and `t.Cleanup()`, and call them at the start of each test that needs them. For example:

```go
func TestThing(t *testing.T) {
	path := setupStuff(t)
	// ...
}

func setupStuff(t *testing.T) string {
	t.Helper()
	t.Cleanup(func() {
		// do any cleanup
	})
	return "/some/path/or/whatever"
}
```

## internal/api/response.go

**Major:** It's usually better for API clients that expect JSON (and I think all of these API endpoints return JSON) to get JSON in the error case as well. You could do this in your `logErrorAndWriteResponse` function:

```go
func logErrorAndWriteResponse(msg string, status int, w http.ResponseWriter) {
```

The name of that is also a bit of a mouthful. Perhaps just `writeError` (logging is an internal detail). I'd also change the signature so you can use `fmt.Sprintf` and formatted string, allowing you to include an error with `%v` easily if needed. It's most common for these functions to take `w` as the first arg. So maybe something like:

```go
func writeError(w http.ResponseWriter, status int, format string, args ...any) {
	type errorResponse struct {
		Error string `json:"error"`
	}

	errorMessage := fmt.Sprintf(format, args...)
	log.Println(errorMessage)

	resp := errorResponse{Error: errorMessage}
	respBytes, err := json.Marshal(&resp)
	if err != nil {
		// log and write a ricsgeneric error and return
	}
	w.WriteHeader(status)
	_, err = w.Write(respBytes)
	if err != nil {
		// log, can't do much else
	}
}
```

## internal/api/middleware.go

**Minor:** `C_STYLE_CONSTS` isn't usually a thing in Go. `USER_ACCOUNT` can be `UserAccount` or perhaps `UserPermission`.

**Minor:** `responseWriterCloner` should probably have an `Unwrap() http.ResponseWriter` method. [Read more about response controllers.](https://www.alexedwards.net/blog/how-to-use-the-http-responsecontroller-type)

**Minor:** It's a bit odd that all the middleware funcs take a generic `ctx *middlewareContext`, instead of just the values they need (for example metrics middleware taking the metrics type).

**Minor:** `responseWriterCloner` is an odd name for this. What about `statusRecorder`?

**Major:** Normally you'd only apply middleware to the handlers that need them. It's messy when you apply them to everything, because then you have to have ad-hoc checks inside the middleware (which should be agnostic to what the route is) like this:

```go
			if !strings.HasPrefix(r.URL.Path, "/_next") {
...
			if !strings.HasPrefix(r.URL.Path, "/api/v1/") {
```

If you apply middleware at the route level (or to a whole group of routes), you can avoid this ad-hoc mathching code in the middleware itself. For example:

```go
	apiMiddleware := createMiddlewareStack(
		authMiddleware(&ctx),
		metricsMiddleware(&ctx),
		loggingMiddleware(&ctx),
	)
	router.Handle("/api/v1/", http.StripPrefix("/api/v1", apiMiddleware(apiV1Router)))

	// For statics, only use metrics middleware
	router.Handle("/", metricsMiddleware(frontendHandler))
```

You should be able to do something similar with the authMiddleware, pass in a "allowFirstAccount" or something, and only wrap the routes you need, to avoid this ad-hoc filtering in the middleware itself:

```
			if r.Method == "POST" && strings.HasSuffix(r.URL.Path, "accounts") && !ctx.firstAccountIssued {
```

Similarly, this looks a bit problematic in `authMiddleware` -- because it means `AllowRequest` has a list of hard-coded routes (done with regex matching):

```
			if claims.Permissions == USER_ACCOUNT {
				requestAllowed, err := AllowRequest(claims, r.Method, r.URL.Path)
```

It would be *much* cleaner to have more specific auth middleware like `adminOnly` and `adminOrUser`, and wrap only the applicable routes in it, for example:

```
	apiV1Router.HandleFunc("GET /accounts", adminOnly(GetUserAccounts(env)))
	apiV1Router.HandleFunc("POST /accounts", adminOnly(PostUserAccount(env)))
	apiV1Router.HandleFunc("DELETE /accounts/{id}", adminOnly(DeleteUserAccount(env)))

	apiV1Router.HandleFunc("GET /accounts/{id}", adminOrUser(GetUserAccount(env)))
	apiV1Router.HandleFunc("POST /accounts/{id}/change_password", adminOrUser(ChangeUserAccountPassword(env)))
```

This would allow you to avoid the kind of re-routing code in `AllowRequest` entirely. You might need to use `Request.PathValue("id")` in `adminOrUser` to get the user ID from the matched URL.

**Major:** Unless I'm misunderstanding, I think this might be a bug (though it'd likely cause a panic in the caller rather than be a security bug). In `getClaimsFromJWT`, if `err` is nil but the token is invalid, this will return `nil, nil` and the caller will assume there's no error:

```go
	if err != nil || !token.Valid {
		return nil, err
	}
```

It should probably be more like:

```go
	if err != nil {
		return nil, err
	}
	if !token.Valid {
		return nil, errors.New("invalid token")
	}
```

## internal/api/authorization_test.go

**Minor:** it'd be cleaner if this special case was a test struct field, `passwordMatch string` or something, instead of a special case that compares the description. (What if someone changes the description and forgets to update this?)

```go
			if tC.desc == "Create no password user success" {
```

Or just pull this out as a separate test function. Not everything needs to be a table-driven test if it's unwieldy.

## internal/api/handlers_certificate_requests.go

**Major::** Generally it's best not to include arbitrary `err.Error()` information in the API responses, in case something mildly sensitive is leaked. For example:

```go
			logErrorAndWriteResponse(err.Error(), http.StatusInternalServerError, w)
```

For internal errors, I would log the full error, but just include something generic like "internal error" in the API response. Also make sure that you have good metrics/alerts on this in production, so you can easily find such errors (which generally "shouldn't happen").

I also find it helpful to make helpers for specific error handling that happens, for example, instead of the above, make a helper you just call as `internalError(w, err)` or `badRequest(w, err)`.

**Major:** You have this "marshal and then write to response" pattern a *lot*:

```go
		body, err := json.Marshal(certs)
		if err != nil {
			logErrorAndWriteResponse(err.Error(), http.StatusInternalServerError, w)
			return
		}
		if _, err := w.Write(body); err != nil {
			logErrorAndWriteResponse(err.Error(), http.StatusInternalServerError, w)
		}
```

It'd be good to extract this to a helper, so you can do this without so much boilerplate:

```go
	err := writeJSON(w, certs)
	if err != nil {
		internalError(w, err)
		return
	}
```

**Major:** API design point. It's usually best to have a "result container" in your API responses, so you can add top-level metadata like "error" or "pagination-cursor" or whatever later. So `writeJSON` could wrap everything in a struct something like so:

```go
type response struct {
	Result any    `json:"result,omitempty"`
}
```

The `writeJSON` function could take care of this wrapping for you, and you could have a `writeJSONError` too that used this shape:

```go
type response struct {
	Error  string `json:"error"`
}
```

**Major:** Testing error strings is really not great, unless you obsoluately have to:

```go
		id, err := env.DB.CreateCSR(string(csr))
		if err != nil {
			if strings.Contains(err.Error(), "UNIQUE constraint failed") {
				logErrorAndWriteResponse("given csr already recorded", http.StatusBadRequest, w)
				return
			}
			if strings.Contains(err.Error(), "csr validation failed") {
				logErrorAndWriteResponse(err.Error(), http.StatusBadRequest, w)
				return
			}
			logErrorAndWriteResponse(err.Error(), http.StatusInternalServerError, w)
			return
		}
```

SQLite error message are probably quite stable (but who knows), but a maintainer might modify the "csv validation failed" error to say "CSV validation failed", and then this check would silently fail.

The `mattn/go-sqlite3` library actually has [good error codes](https://github.com/mattn/go-sqlite3/blob/846fea6c1443e8cc366fc1966fe078d7f825f6a9/error.go#L144) for things like this. You'd check with:

```go
if errors.Is(err, sqlite3.ErrConstraintUnique) {
	...
}
```

However, it's probably best to not let lower-level `sqlite3` errors leak through the database package to the api layer. You probably want to do that check in the db package and convert it to a custom error type like `db.ErrAlreadyExists`, defined at the package level like so:

```go
var ErrAlreadyExists = errors.New("already exists")
```

Oh, I see you already have an `ErrIdNotFound` -- yeah, use exactly that same pattern here -- you want to avoid error string matching like the plague.

For the "csr validation error", you can do the same thing:

```go
	// in db.go:
	var ErrCSRValidationFailed = errors.New("csr validation failed")
	// ...
	if err := ValidateCertificateRequest(csr); err != nil {
		return 0, ErrCSRValidationFailed
	}

	// in handler:
	if errors.Is(err, db.ErrCSRValidationFailed) {
		logErrorAndWriteResponse(err.Error(), http.StatusBadRequest, w)
		return
	}
```

If you're finding all the database error checking is basically the same ... in one of my projects I have a helper `handleDatabaseError` that does all the different checks and returns the correct error (bad request, internal error, etc). Something like so:

```go
// handleDatabaseError checks for specific database errors and writes the appropriate response,
// returning true if it is an error, otherwise false.
func handleDatabaseError(w http.ResponseWriter, err error) bool {
	if errors.Is(err, db.NotFound) {
		badRequest(w, err)
		return true
	}
	if err != nil {
		internalError(w, err)
		return true
	}
	return false
}

// use like so:
err := db.SomeDBOperation()
if handleDatabaseError(w, err) {
	return
}
```

Which avoids a lot of boilerplate after database queries.

**Major:** As a consumer of the API, it's weird that some things return JSON, but `PostCertificateRequest` returns a raw integer:

```go
		if _, err := w.Write([]byte(strconv.FormatInt(id, 10))); err != nil {
			logErrorAndWriteResponse(err.Error(), http.StatusInternalServerError, w)
		}
```

In this case `strconv.FormatInt(id, 10)` happens to be in JSON number format, but that's kind of by coincidence.

I think all responses should be wrapped in `JSON`. It's usually best to have a standard "shape", for example `{"result": ...}` -- see my comments about `writeJSON` above, and then you could use just `writeJSON(w, id)`.

**Minor:** I know it's only a pseudo-domain, but you should probably use a "domain" that we own in the Pebble notify key:

```go
err := SendPebbleNotification("notary.com/certificate/update", insertIdStr)
```

Maybe:

```go
err := SendPebbleNotification("canonical.com/notary/certificate/update", insertIdStr)
```

**Minor:** Speaking of pebble notify (it's cool that you're using it BTW). If `conf.PebbleNotificationsEnabled`, you should probably check at startup that `exec.LookPath` can find the `pebble` binary, so it doesn't fail much later at runtime and you potentially miss the failures. It seems to me if that's turned on it should be a hard failure.

**Major:** It looks like it sends a "certificate/update" pebble notify even for the reject and delete cases. Shouldn't these be a different notice key?

## internal/api/handlers_certificate_requests_test.go

**Minor:** For things like this:

```go
expectedGetAllCertsResponseBody1 = fmt.Sprintf("[{\"id\":1,\"csr\":\"%s\",\"certificate\":\"\"}]", trimmed(AppleCSR))
```

I would use a little helper (you can run helpers at the package-level for variable initialisation) that actually used `json.Marshal` with a little struct type. It would make this safer/cleaner (what if some of the strings needs JSON-escaping?).

## internal/api/handlers_health.go

**Minor:** In `HealthCheck`, you have to write the headers/status before the content, so swap these two lines:

```go
		w.Write(response)            //nolint:errcheck
		w.WriteHeader(http.StatusOK) //nolint:errcheck
```

It also wouldn't hurt to log the errors and avoid the "nolint" comments.

## internal/api/handlers_login.go:

**Minor:** This literal is a bit hidden away in the code:

```go
			ExpiresAt: time.Now().Add(time.Hour * 1).Unix(),
```

Might be nice to make a package-level named `const`.

## internal/api/handlers_login_test.go

**Minor:** Once again, better to do this kind of thing with a struct field in the test case, rather than ad-hoc string matching on the test description as a special case:

```go
			if tC.desc == "Login success" && res.StatusCode == http.StatusOK {
```

Or just pull this out as a separate test function. Not everything needs to be a table-driven test if it's unwieldy.

kayra1:
# Description

This change addresses the final points and closes #84, which modifies the database to have explicit ID's, adds a new status column and associated SQL checks, and utilizes SQLair to simplify database SQL calls.

It also changes the API of the database to fit the new way of using the status column. We now have functions that are abstracted from the direct SQL calls, like `UpdateCSR` -> `AddCertificateToCSR`. This is to simplify future development and help avoid getting the database into inconsistent states.

Another change that SQLair brings is the fact that it no longer returns the Result and LastID values that we used to pass along to the HTTP server. This means that successful API calls only receive a JSON message indicating success instead.

# Checklist:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,6 +12,7 @@ require (
 
 require (
 	github.com/beorn7/perks v1.0.1 // indirect
+	github.com/canonical/sqlair v0.0.0-20241004123011-77313b5382fd
 	github.com/cespare/xxhash/v2 v2.3.0 // indirect
 	github.com/klauspost/compress v1.17.9 // indirect
 	github.com/kr/text v0.2.0 // indirect
--- a/internal/db/db.go
+++ b/internal/db/db.go
@@ -24,7 +24,7 @@ type CertificateRequest struct {
 
 	CSR              string `db:"csr"`
 	CertificateChain string `db:"certificate_chain"`
-	RequestStatus    string `db:"request_status"`
+	Status           string `db:"status"`
 }
 
 type User struct {
@@ -46,7 +46,7 @@ const queryCreateCertificateRequestsTable = `
 
 		csr TEXT NOT NULL UNIQUE, 
 		certificate_chain TEXT DEFAULT '',
-		request_status TEXT DEFAULT 'Outstanding', 
+		status TEXT DEFAULT 'Outstanding', 
 		
 		CHECK (request_status IN ('Outstanding', 'Rejected', 'Revoked', 'Active')),
 		CHECK (NOT (certificate_chain == '' AND request_status == 'Active' )),
@@ -65,15 +65,15 @@ const queryCreateUsersTable = `
 )`
 
 const (
-	getAllCSRsStmt = "SELECT &CertificateRequest.* FROM %s"
-	getCSRsStmt    = "SELECT &CertificateRequest.* FROM %s WHERE id==$CertificateRequest.id or csr==$CertificateRequest.csr"
-	createCSRStmt  = "INSERT INTO %s (csr) VALUES ($CertificateRequest.csr)"
-	updateCSRStmt  = "UPDATE %s SET certificate_chain=$CertificateRequest.certificate_chain, request_status=$CertificateRequest.request_status WHERE id==$CertificateRequest.id or csr==$CertificateRequest.csr"
-	deleteCSRStmt  = "DELETE FROM %s WHERE id=$CertificateRequest.id or csr=$CertificateRequest.csr"
+	listCertificateRequestsStmt  = "SELECT &CertificateRequest.* FROM %s"
+	getCertificateRequestStmt    = "SELECT &CertificateRequest.* FROM %s WHERE id==$CertificateRequest.id or csr==$CertificateRequest.csr"
+	createCertificateRequestStmt = "INSERT INTO %s (csr) VALUES ($CertificateRequest.csr)"
+	updateCertificateRequestStmt = "UPDATE %s SET certificate_chain=$CertificateRequest.certificate_chain, request_status=$CertificateRequest.request_status WHERE id==$CertificateRequest.id or csr==$CertificateRequest.csr"
+	deleteCertificateRequestStmt = "DELETE FROM %s WHERE id=$CertificateRequest.id or csr=$CertificateRequest.csr"
 )
 
 const (
-	getAllUsersStmt = "SELECT &User.* from %s"
+	listUsersStmt   = "SELECT &User.* from %s"
 	getUserStmt     = "SELECT &User.* from %s WHERE id==$User.id or username==$User.username"
 	createUserStmt  = "INSERT INTO %s (username, hashed_password, permissions) VALUES ($User.username, $User.hashed_password, $User.permissions)"
 	updateUserStmt  = "UPDATE %s SET hashed_password=$User.hashed_password WHERE id==$User.id or username==$User.username"
@@ -81,9 +81,9 @@ const (
 	getNumUsersStmt = "SELECT COUNT(*) AS &NumUsers.count FROM %s"
 )
 
-// RetrieveAllCSRs gets every CertificateRequest entry in the table.
-func (db *Database) RetrieveAllCSRs() ([]CertificateRequest, error) {
-	stmt, err := sqlair.Prepare(fmt.Sprintf(getAllCSRsStmt, db.certificateTable), CertificateRequest{})
+// ListCertificateRequests gets every CertificateRequest entry in the table.
+func (db *Database) ListCertificateRequests() ([]CertificateRequest, error) {
+	stmt, err := sqlair.Prepare(fmt.Sprintf(listCertificateRequestsStmt, db.certificateTable), CertificateRequest{})
 	if err != nil {
 		return nil, err
 	}
@@ -98,12 +98,12 @@ func (db *Database) RetrieveAllCSRs() ([]CertificateRequest, error) {
 	return csrs, nil
 }
 
-// RetrieveCSRbyID gets a CSR row from the repository from a given ID.
-func (db *Database) RetrieveCSRbyID(id int) (*CertificateRequest, error) {
+// GetCertificateRequestByID gets a CSR row from the repository from a given ID.
+func (db *Database) GetCertificateRequestByID(id int) (*CertificateRequest, error) {
 	csr := CertificateRequest{
 		ID: id,
 	}
-	stmt, err := sqlair.Prepare(fmt.Sprintf(getCSRsStmt, db.certificateTable), CertificateRequest{})
+	stmt, err := sqlair.Prepare(fmt.Sprintf(getCertificateRequestStmt, db.certificateTable), CertificateRequest{})
 	if err != nil {
 		return nil, err
 	}
@@ -114,12 +114,12 @@ func (db *Database) RetrieveCSRbyID(id int) (*CertificateRequest, error) {
 	return &csr, nil
 }
 
-// RetrieveCSRbyCSR gets a given CSR row from the repository using the CSR text.
-func (db *Database) RetrieveCSRbyCSR(csr string) (*CertificateRequest, error) {
+// GetCertificateRequestByCSR gets a given CSR row from the repository using the CSR text.
+func (db *Database) GetCertificateRequestByCSR(csr string) (*CertificateRequest, error) {
 	row := CertificateRequest{
 		CSR: csr,
 	}
-	stmt, err := sqlair.Prepare(fmt.Sprintf(getCSRsStmt, db.certificateTable), CertificateRequest{})
+	stmt, err := sqlair.Prepare(fmt.Sprintf(getCertificateRequestStmt, db.certificateTable), CertificateRequest{})
 	if err != nil {
 		return nil, err
 	}
@@ -130,12 +130,12 @@ func (db *Database) RetrieveCSRbyCSR(csr string) (*CertificateRequest, error) {
 	return &row, nil
 }
 
-// CreateCSR creates a new CSR entry in the repository. The string must be a valid CSR and unique.
-func (db *Database) CreateCSR(csr string) error {
+// CreateCertificateRequest creates a new CSR entry in the repository. The string must be a valid CSR and unique.
+func (db *Database) CreateCertificateRequest(csr string) error {
 	if err := ValidateCertificateRequest(csr); err != nil {
 		return errors.New("csr validation failed: " + err.Error())
 	}
-	stmt, err := sqlair.Prepare(fmt.Sprintf(createCSRStmt, db.certificateTable), CertificateRequest{})
+	stmt, err := sqlair.Prepare(fmt.Sprintf(createCertificateRequestStmt, db.certificateTable), CertificateRequest{})
 	if err != nil {
 		return err
 	}
@@ -146,8 +146,8 @@ func (db *Database) CreateCSR(csr string) error {
 	return err
 }
 
-// AddCertificateChainToCSRbyCSR adds a new certificate chain to a row for a given CSR string.
-func (db *Database) AddCertificateChainToCSRbyCSR(csr string, cert string) error {
+// AddCertificateChainToCertificateRequestByCSR adds a new certificate chain to a row for a given CSR string.
+func (db *Database) AddCertificateChainToCertificateRequestByCSR(csr string, cert string) error {
 	err := ValidateCertificate(cert)
 	if err != nil {
 		return errors.New("cert validation failed: " + err.Error())
@@ -157,22 +157,22 @@ func (db *Database) AddCertificateChainToCSRbyCSR(csr string, cert string) error
 		return errors.New("cert validation failed: " + err.Error())
 	}
 	certBundle := sanitizeCertificateBundle(cert)
-	stmt, err := sqlair.Prepare(fmt.Sprintf(updateCSRStmt, db.certificateTable), CertificateRequest{})
+	stmt, err := sqlair.Prepare(fmt.Sprintf(updateCertificateRequestStmt, db.certificateTable), CertificateRequest{})
 	if err != nil {
 		return err
 	}
 	newRow := CertificateRequest{
 		CSR:              csr,
 		CertificateChain: certBundle,
-		RequestStatus:    "Active",
+		Status:           "Active",
 	}
 	err = db.conn.Query(context.Background(), stmt, newRow).Run()
 	return err
 }
 
 // AddCertificateChainToCSRbyID adds a new certificate chain to a row for a given row ID.
-func (db *Database) AddCertificateToCSRbyID(id int, cert string) error {
-	csr, err := db.RetrieveCSRbyID(id)
+func (db *Database) AddCertificateChainToCertificateRequestByID(id int, cert string) error {
+	csr, err := db.GetCertificateRequestByID(id)
 	if err != nil {
 		return err
 	}
@@ -185,26 +185,26 @@ func (db *Database) AddCertificateToCSRbyID(id int, cert string) error {
 		return errors.New("cert validation failed: " + err.Error())
 	}
 	certBundle := sanitizeCertificateBundle(cert)
-	stmt, err := sqlair.Prepare(fmt.Sprintf(updateCSRStmt, db.certificateTable), CertificateRequest{})
+	stmt, err := sqlair.Prepare(fmt.Sprintf(updateCertificateRequestStmt, db.certificateTable), CertificateRequest{})
 	if err != nil {
 		return err
 	}
 	newRow := CertificateRequest{
 		ID:               id,
 		CertificateChain: certBundle,
-		RequestStatus:    "Active",
+		Status:           "Active",
 	}
 	err = db.conn.Query(context.Background(), stmt, newRow).Run()
 	return err
 }
 
-// RejectCSRbyCSR updates input CSR's row by setting the certificate bundle to "" and moving the row status to "Rejected".
-func (db *Database) RejectCSRbyCSR(csr string) error {
-	oldRow, err := db.RetrieveCSRbyCSR(csr)
+// RejectCertificateRequestByCSR updates input CSR's row by setting the certificate bundle to "" and moving the row status to "Rejected".
+func (db *Database) RejectCertificateRequestByCSR(csr string) error {
+	oldRow, err := db.GetCertificateRequestByCSR(csr)
 	if err != nil {
 		return err
 	}
-	stmt, err := sqlair.Prepare(fmt.Sprintf(updateCSRStmt, db.certificateTable), CertificateRequest{})
+	stmt, err := sqlair.Prepare(fmt.Sprintf(updateCertificateRequestStmt, db.certificateTable), CertificateRequest{})
 	if err != nil {
 		return err
 	}
@@ -212,39 +212,19 @@ func (db *Database) RejectCSRbyCSR(csr string) error {
 		ID:               oldRow.ID,
 		CSR:              oldRow.CSR,
 		CertificateChain: "",
-		RequestStatus:    "Rejected",
+		Status:           "Rejected",
 	}
 	err = db.conn.Query(context.Background(), stmt, newRow).Run()
 	return err
 }
 
 // RejectCSRbyCSR updates input ID's row by setting the certificate bundle to "" and sets the row status to "Rejected".
-func (db *Database) RejectCSRbyID(id int) error {
-	oldRow, err := db.RetrieveCSRbyID(id)
+func (db *Database) RejectCertificateRequestByID(id int) error {
+	oldRow, err := db.GetCertificateRequestByID(id)
 	if err != nil {
 		return err
 	}
-	stmt, err := sqlair.Prepare(fmt.Sprintf(updateCSRStmt, db.certificateTable), CertificateRequest{})
-	if err != nil {
-		return err
-	}
-	newRow := CertificateRequest{
-		ID:               oldRow.ID,
-		CSR:              oldRow.CSR,
-		CertificateChain: "",
-		RequestStatus:    "Rejected",
-	}
-	err = db.conn.Query(context.Background(), stmt, newRow).Run()
-	return err
-}
-
-// RevokeCSR updates the input CSR's row by setting the certificate bundle to "" and sets the row status to "Revoked".
-func (db *Database) RevokeCSR(csr string) error {
-	oldRow, err := db.RetrieveCSRbyCSR(csr)
-	if err != nil {
-		return err
-	}
-	stmt, err := sqlair.Prepare(fmt.Sprintf(updateCSRStmt, db.certificateTable), CertificateRequest{})
+	stmt, err := sqlair.Prepare(fmt.Sprintf(updateCertificateRequestStmt, db.certificateTable), CertificateRequest{})
 	if err != nil {
 		return err
 	}
@@ -252,15 +232,35 @@ func (db *Database) RevokeCSR(csr string) error {
 		ID:               oldRow.ID,
 		CSR:              oldRow.CSR,
 		CertificateChain: "",
-		RequestStatus:    "Revoked",
+		Status:           "Rejected",
 	}
 	err = db.conn.Query(context.Background(), stmt, newRow).Run()
 	return err
 }
 
-// DeleteCSRbyCSR removes a CSR from the database alongside the certificate that may have been generated for it.
-func (db *Database) DeleteCSRbyCSR(csr string) error {
-	stmt, err := sqlair.Prepare(fmt.Sprintf(deleteCSRStmt, db.certificateTable), CertificateRequest{})
+// RevokeCertificateByCSR updates the input CSR's row by setting the certificate bundle to "" and sets the row status to "Revoked".
+func (db *Database) RevokeCertificateByCSR(csr string) error {
+	oldRow, err := db.GetCertificateRequestByCSR(csr)
+	if err != nil {
+		return err
+	}
+	stmt, err := sqlair.Prepare(fmt.Sprintf(updateCertificateRequestStmt, db.certificateTable), CertificateRequest{})
+	if err != nil {
+		return err
+	}
+	newRow := CertificateRequest{
+		ID:               oldRow.ID,
+		CSR:              oldRow.CSR,
+		CertificateChain: "",
+		Status:           "Revoked",
+	}
+	err = db.conn.Query(context.Background(), stmt, newRow).Run()
+	return err
+}
+
+// DeleteCertificateRequestByCSR removes a CSR from the database alongside the certificate that may have been generated for it.
+func (db *Database) DeleteCertificateRequestByCSR(csr string) error {
+	stmt, err := sqlair.Prepare(fmt.Sprintf(deleteCertificateRequestStmt, db.certificateTable), CertificateRequest{})
 	if err != nil {
 		return err
 	}
@@ -272,8 +272,8 @@ func (db *Database) DeleteCSRbyCSR(csr string) error {
 }
 
 // DeleteCSRByID removes a CSR from the database alongside the certificate that may have been generated for it.
-func (db *Database) DeleteCSRbyID(id int) error {
-	stmt, err := sqlair.Prepare(fmt.Sprintf(deleteCSRStmt, db.certificateTable), CertificateRequest{})
+func (db *Database) DeleteCertificateRequestByID(id int) error {
+	stmt, err := sqlair.Prepare(fmt.Sprintf(deleteCertificateRequestStmt, db.certificateTable), CertificateRequest{})
 	if err != nil {
 		return err
 	}
@@ -284,9 +284,9 @@ func (db *Database) DeleteCSRbyID(id int) error {
 	return err
 }
 
-// RetrieveAllUsers returns all of the users and their fields available in the database.
-func (db *Database) RetrieveAllUsers() ([]User, error) {
-	stmt, err := sqlair.Prepare(fmt.Sprintf(getAllUsersStmt, db.usersTable), User{})
+// ListUsers returns all of the users and their fields available in the database.
+func (db *Database) ListUsers() ([]User, error) {
+	stmt, err := sqlair.Prepare(fmt.Sprintf(listUsersStmt, db.usersTable), User{})
 	if err != nil {
 		return nil, err
 	}
@@ -298,8 +298,8 @@ func (db *Database) RetrieveAllUsers() ([]User, error) {
 	return users, nil
 }
 
-// RetrieveUser retrieves the name, password and the permission level of a user.
-func (db *Database) RetrieveUserByID(id int) (*User, error) {
+// GetUserByID retrieves the name, password and the permission level of a user.
+func (db *Database) GetUserByID(id int) (*User, error) {
 	row := User{
 		ID: id,
 	}
@@ -314,8 +314,8 @@ func (db *Database) RetrieveUserByID(id int) (*User, error) {
 	return &row, nil
 }
 
-// RetrieveUser retrieves the id, password and the permission level of a user.
-func (db *Database) RetrieveUserByUsername(name string) (*User, error) {
+// GetUserByUsername retrieves the id, password and the permission level of a user.
+func (db *Database) GetUserByUsername(name string) (*User, error) {
 	row := User{
 		Username: name,
 	}
@@ -354,7 +354,7 @@ func (db *Database) CreateUser(username string, password string, permission int)
 // UpdateUser updates the password of the given user.
 // Just like with CreateUser, this function handles hashing and salting the password before storage.
 func (db *Database) UpdateUserPassword(id int, password string) error {
-	_, err := db.RetrieveUserByID(id)
+	_, err := db.GetUserByID(id)
 	if err != nil {
 		return err
 	}
@@ -376,7 +376,7 @@ func (db *Database) UpdateUserPassword(id int, password string) error {
 
 // DeleteUserByID removes a user from the table.
 func (db *Database) DeleteUserByID(id int) error {
-	_, err := db.RetrieveUserByID(id)
+	_, err := db.GetUserByID(id)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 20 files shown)


go.mod
```diff
@@ -12,6 +12,7 @@ require (
 
 require (
 	github.com/beorn7/perks v1.0.1 // indirect
+	github.com/canonical/sqlair v0.0.0-20241004123011-77313b5382fd
```
gruyaume, line 15: It's strange that they don't really version sqlair.

internal/db/db.go
```diff
@@ -2,281 +2,452 @@
...
 
-const queryCreateCSRsTable = `CREATE TABLE IF NOT EXISTS %s (
-	csr TEXT PRIMARY KEY UNIQUE NOT NULL, 
-	certificate TEXT DEFAULT ''
+const queryCreateCSRsTable = `
+	CREATE TABLE IF NOT EXISTS %s (
+	    id INTEGER PRIMARY KEY AUTOINCREMENT,
+
+		csr TEXT NOT NULL UNIQUE, 
+		certificate_chain TEXT DEFAULT '',
```
gruyaume, line 48: Why have a default value of an empty string?

internal/db/db.go
```diff
@@ -2,281 +2,452 @@
...
 )`
 
-const (
-	queryGetAllCSRs = "SELECT rowid, * FROM %s"
-	queryGetCSR     = "SELECT rowid, * FROM %s WHERE rowid=?"
-	queryCreateCSR  = "INSERT INTO %s (csr) VALUES (?)"
-	queryUpdateCSR  = "UPDATE %s SET certificate=? WHERE rowid=?"
-	queryDeleteCSR  = "DELETE FROM %s WHERE rowid=?"
-)
+const queryCreateUsersTable = `
```
gruyaume, line 58: (Unrelated to this PR - feel free to ignore) 

We'll likely want to relate users with certificate requests so that it's clear which user made which request

internal/db/db.go
```diff
@@ -2,281 +2,452 @@
...
 const (
-	certificateRequestsTableName = "CertificateRequests"
+	certificateRequestsTableName = "certificate_requests"
 	usersTableName               = "users"
 )
 
-const queryCreateCSRsTable = `CREATE TABLE IF NOT EXISTS %s (
-	csr TEXT PRIMARY KEY UNIQUE NOT NULL, 
-	certificate TEXT DEFAULT ''
+const queryCreateCSRsTable = `
```
gruyaume, line 43:
```suggestion
const queryCreateCertificateRequestsTable = `
```

internal/db/db.go
```diff
@@ -2,281 +2,452 @@
...
+	return &csr, nil
 }
-type User struct {
-	ID          int
-	Username    string
-	Password    string
-	Permissions int
+
+// RetrieveCSRbyCSR gets a given CSR row from the repository using the CSR text.
+func (db *Database) RetrieveCSRbyCSR(csr string) (*CertificateRequest, error) {
```
gruyaume, line 118:
```suggestion
func (db *Database) RetrieveCertificateRequestByCSR(csr string) (*CertificateRequest, error) {
```

internal/db/db.go
```diff
@@ -2,281 +2,452 @@
...
+	err = db.conn.Query(context.Background(), stmt, row).Get(&row)
+	if err != nil {
+		return nil, err
+	}
+	return &row, nil
 }
 
-var ErrIdNotFound = errors.New("id not found")
+// CreateCSR creates a new CSR entry in the repository. The string must be a valid CSR and unique.
+func (db *Database) CreateCSR(csr string) error {
```
gruyaume, line 134:
```suggestion
func (db *Database) CreateCertificateRequest(csr string) error {
```

gruyaume: Generally speaking, I think we abuse the term CSR a bit here and in other places, I made a couple of suggestions to improve this.

internal/db/db.go
```diff
@@ -2,281 +2,452 @@
...
+		return err
+	}
+	return nil
+}
 
-// RetrieveAllCSRs gets every CertificateRequest entry in the table.
-func (db *Database) RetrieveAllCSRs() ([]CertificateRequest, error) {
-	rows, err := db.conn.Query(fmt.Sprintf(queryGetAllCSRs, db.certificateTable))
+// AddCertificateChainToCSRbyCSR adds a new certificate chain to a row for a given CSR string.
+func (db *Database) AddCertificateChainToCSRbyCSR(csr string, cert string) error {
```
gruyaume, line 153:
```suggestion
func (db *Database) AddCertificateChainToCertificateRequestByCSR(csr string, cert string) error {
```

internal/db/db.go
```diff
@@ -2,281 +2,452 @@
...
-	var allCsrs []CertificateRequest
-	defer rows.Close()
-	for rows.Next() {
-		var csr CertificateRequest
-		if err := rows.Scan(&csr.ID, &csr.CSR, &csr.Certificate); err != nil {
-			return nil, err
-		}
-		allCsrs = append(allCsrs, csr)
+// AddCertificateChainToCSRbyID adds a new certificate chain to a row for a given row ID.
+func (db *Database) AddCertificateToCSRbyID(id int, cert string) error {
```
gruyaume, line 180:
```suggestion
func (db *Database) AddCertificateToCertificateRequestByID(id int, cert string) error {
```

gruyaume, line 180: Do we use this anywhere?

internal/db/db.go
```diff
@@ -2,281 +2,452 @@
...
-func (db *Database) RetrieveCSR(id string) (CertificateRequest, error) {
-	var newCSR CertificateRequest
-	row := db.conn.QueryRow(fmt.Sprintf(queryGetCSR, db.certificateTable), id)
-	if err := row.Scan(&newCSR.ID, &newCSR.CSR, &newCSR.Certificate); err != nil {
-		if err.Error() == "sql: no rows in result set" {
-			return newCSR, ErrIdNotFound
-		}
-		return newCSR, err
+// RejectCSRbyCSR updates input CSR's row by setting the certificate bundle to "" and moving the row status to "Rejected".
+func (db *Database) RejectCSRbyCSR(csr string) error {
```
gruyaume, line 211:
```suggestion
func (db *Database) RejectCertificateRequestByCSR(csr string) error {
```

internal/db/db.go
```diff
@@ -2,281 +2,452 @@
...
 }
 
-// A CertificateRequest struct represents an entry in the database.
-// The object contains a Certificate Request, its matching Certificate if any, and the row ID.
-type CertificateRequest struct {
-	ID          int
-	CSR         string
-	Certificate string
+// RetrieveCSRbyID gets a CSR row from the repository from a given ID.
+func (db *Database) RetrieveCSRbyID(id int) (*CertificateRequest, error) {
```
gruyaume, line 102:
```suggestion
func (db *Database) RetrieveCertificateRequestByID(id int) (*CertificateRequest, error) {
```

internal/db/db.go
```diff
@@ -2,281 +2,452 @@
...
+	getAllUsersStmt = "SELECT &User.* from %s"
+	getUserStmt     = "SELECT &User.* from %s WHERE id==$User.id or username==$User.username"
+	createUserStmt  = "INSERT INTO %s (username, hashed_password, permissions) VALUES ($User.username, $User.hashed_password, $User.permissions)"
+	updateUserStmt  = "UPDATE %s SET hashed_password=$User.hashed_password WHERE id==$User.id or username==$User.username"
+	deleteUserStmt  = "DELETE FROM %s WHERE id==$User.id"
+	getNumUsersStmt = "SELECT COUNT(*) AS &NumUsers.count FROM %s"
+)
+
+// RetrieveAllCSRs gets every CertificateRequest entry in the table.
+func (db *Database) RetrieveAllCSRs() ([]CertificateRequest, error) {
```
gruyaume, line 85:
```suggestion
func (db *Database) RetrieveAllCertificateRequests() ([]CertificateRequest, error) {
```

internal/db/db.go
```diff
@@ -2,281 +2,452 @@
...
-	return id, nil
+	return nil
 }
 
-// UpdateCSR adds a new cert to the given CSR in the repository.
-// The given certificate must share the public key of the CSR and must be valid.
-func (db *Database) UpdateCSR(id string, cert string) (int64, error) {
-	csr, err := db.RetrieveCSR(id)
+// RevokeCSR updates the input CSR's row by setting the certificate bundle to "" and sets the row status to "Revoked".
+func (db *Database) RevokeCSR(csr string) error {
```
gruyaume, line 257:
```suggestion
func (db *Database) RevokeCertificateRequest(csr string) error {
```

internal/db/db.go
```diff
@@ -2,281 +2,452 @@
...
 	}
-	return int64(csr.ID), nil
+	return nil
 }
 
-// DeleteCSR removes a CSR from the database alongside the certificate that may have been generated for it.
-func (db *Database) DeleteCSR(id string) (int64, error) {
-	result, err := db.conn.Exec(fmt.Sprintf(queryDeleteCSR, db.certificateTable), id)
+// DeleteCSRbyCSR removes a CSR from the database alongside the certificate that may have been generated for it.
+func (db *Database) DeleteCSRbyCSR(csr string) error {
```
gruyaume, line 280:
```suggestion
func (db *Database) DeleteCertificateRequestByCSR(csr string) error {
```

internal/db/db.go
```diff
@@ -2,281 +2,452 @@
...
-		return 0, err
+		return err
 	}
-	if deleteId == 0 {
-		return 0, ErrIdNotFound
+	return nil
+}
+
+// DeleteCSRByID removes a CSR from the database alongside the certificate that may have been generated for it.
+func (db *Database) DeleteCSRbyID(id int) error {
```
gruyaume, line 296:
```suggestion
func (db *Database) DeleteCertificateRequestByID(id int) error {
```

internal/server/handlers_certificate_requests_test.go
```diff
@@ -9,22 +9,24 @@ import (
...
-type CertificateRequest struct {
-	ID          int    `json:"id"`
-	CSR         string `json:"csr"`
-	Certificate string `json:"certificate"`
-}
+// type CertificateRequest struct {
+// 	ID          int    `json:"id"`
+// 	CSR         string `json:"csr"`
+// 	Certificate string `json:"certificate"`
+// }
```
gruyaume, line 20: Should we get rid of this?

internal/server/handlers_certificate_requests.go
```diff
@@ -8,62 +8,40 @@ import (
...
-	ID int `json:"id"`
-}
-
-type RejectCertificateResponse struct {
-	ID int `json:"id"`
+type CertificateRequest struct {
+	ID               int    `json:"id"`
+	CSR              string `json:"csr"`
+	CertificateChain string `json:"certificate_chain"`
+	CSRStatus        string `json:"csr_status"`
```
gruyaume, line 26:
```suggestion
	status        string `json:"status"`
```

gruyaume (changes requested):
We use the word `CSR` to mean two different things here, I made a couple of suggestions to decouple CertificateRequest from CSR, where CSR is the X.509 document and a CertificateRequest is an object that contains the CSR, the Certificate and the Status.

internal/db/db.go
```diff
@@ -2,281 +2,452 @@
...
+	usersTable       string
+	conn             *sqlair.DB
+}
+
+type CertificateRequest struct {
+	ID int `db:"id"`
+
+	CSR              string `db:"csr"`
+	CertificateChain string `db:"certificate_chain"`
+	RequestStatus    string `db:"request_status"`
```
gruyaume, line 27:
```suggestion
	Status    string `db:"status"`
```

internal/db/db.go
```diff
@@ -2,281 +2,452 @@
...
-const queryCreateCSRsTable = `CREATE TABLE IF NOT EXISTS %s (
-	csr TEXT PRIMARY KEY UNIQUE NOT NULL, 
-	certificate TEXT DEFAULT ''
+const queryCreateCSRsTable = `
+	CREATE TABLE IF NOT EXISTS %s (
+	    id INTEGER PRIMARY KEY AUTOINCREMENT,
+
+		csr TEXT NOT NULL UNIQUE, 
+		certificate_chain TEXT DEFAULT '',
+		request_status TEXT DEFAULT 'Outstanding',
```
gruyaume, line 49:
```suggestion
		status TEXT DEFAULT 'Outstanding', 
```

gruyaume (changes requested):

internal/db/db.go
```diff
@@ -2,281 +2,452 @@
...
 
 const (
-	queryGetAllUsers       = "SELECT * FROM %s"
-	queryGetUser           = "SELECT * FROM %s WHERE user_id=?"
-	queryGetUserByUsername = "SELECT * FROM %s WHERE username=?"
-	queryCreateUser        = "INSERT INTO %s (username, hashed_password, permissions) VALUES (?, ?, ?)"
-	queryUpdateUser        = "UPDATE %s SET hashed_password=? WHERE user_id=?"
-	queryDeleteUser        = "DELETE FROM %s WHERE user_id=?"
-	queryGetNumUsers       = "SELECT COUNT(*) FROM %s"
+	getAllCSRsStmt = "SELECT &CertificateRequest.* FROM %s"
```
gruyaume, line 68:
```suggestion
	listCertificateRequestsStmt = "SELECT &CertificateRequest.* FROM %s"
```

internal/db/db.go
```diff
@@ -2,281 +2,452 @@
...
+	getCSRsStmt    = "SELECT &CertificateRequest.* FROM %s WHERE id==$CertificateRequest.id or csr==$CertificateRequest.csr"
```
gruyaume, line 69:
```suggestion
	getCertificateRequestStmt    = "SELECT &CertificateRequest.* FROM %s WHERE id==$CertificateRequest.id or csr==$CertificateRequest.csr"
```

internal/db/db.go
```diff
@@ -2,281 +2,452 @@
...
+	createCSRStmt  = "INSERT INTO %s (csr) VALUES ($CertificateRequest.csr)"
```
gruyaume, line 70:
```suggestion
	createCertificateRequestStmt  = "INSERT INTO %s (csr) VALUES ($CertificateRequest.csr)"
```

internal/db/db.go
```diff
@@ -2,281 +2,452 @@
...
+	updateCSRStmt  = "UPDATE %s SET certificate_chain=$CertificateRequest.certificate_chain, request_status=$CertificateRequest.request_status WHERE id==$CertificateRequest.id or csr==$CertificateRequest.csr"
```
gruyaume, line 71:
```suggestion
	updateCertificateRequestStmt  = "UPDATE %s SET certificate_chain=$CertificateRequest.certificate_chain, request_status=$CertificateRequest.request_status WHERE id==$CertificateRequest.id or csr==$CertificateRequest.csr"
```

internal/db/db.go
```diff
@@ -2,281 +2,452 @@
...
+	deleteCSRStmt  = "DELETE FROM %s WHERE id=$CertificateRequest.id or csr=$CertificateRequest.csr"
```
gruyaume, line 72:
```suggestion
	deleteCertificateRequestStmt  = "DELETE FROM %s WHERE id=$CertificateRequest.id or csr=$CertificateRequest.csr"
```

gruyaume (changes requested):

internal/metrics/metrics.go
```diff
@@ -218,6 +218,7 @@ func requestDurationMetric() prometheus.HistogramVec {
 }
 
 func certificateExpiryDate(certString string) time.Time {
+	// TODO: Does this return the expiry date of the issuer?
```
saltiyazan, line 221: I like avoiding TODO comments. I'm not sure what's the goal of this one here anyway.
If we are passing the chain to the function, I think this function would return the expiry date of the first cert in the chain.
Should we create a bug or fix it here instead of keeping the TODO?

saltiyazan (changes requested):
In general this looks good.

I added a few minor comments, and I agree with Guillaume that it makes it easier to read if the Request object and the CSR string have different names.

go.mod
kayra1: :man_shrugging:

internal/db/db.go
kayra1: done

internal/db/db.go
kayra1: yes, in the certificate request handler tests

internal/db/db.go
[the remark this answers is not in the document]
kayra1: Retrieve was for database queries, get was for handler functions. updated.

ui/src/app/(notary)/certificate_requests/table.tsx
```diff
@@ -185,13 +185,13 @@ export function CertificateRequestsTable({ csrs: rows }: TableProps) {
...
+                                        {isCSRContentVisible ? "Hide Certificate Request Content" : "Show Certificate Request Content"}
                                     </Button>
                                     <Button
                                         className="p-contextual-menu__link"
-                                        disabled={certificate == "rejected"}
-                                        onMouseDown={() => handleReject(id)}>
-                                        Reject Certificate Request
+                                        disabled={csr_status == "Rejected"}
+                                        onMouseDown={() => csr_status == "Active" ? handleRevoke(id) : handleReject(id)}>
+                                        {csr_status == "Active" ? "Revoke Certificate Request" : "Reject Certificate Request"}
```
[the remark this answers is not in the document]
kayra1: we are actually rejecting the request though aren't we?

ui/src/types.ts
```diff
@@ -1,7 +1,8 @@
 export type CSREntry = {
     id: number,
     csr: string,
-    certificate: string
+    certificate_chain: string
+    csr_status: "Outstanding" | "Active" | "Rejected" | "Revoked"
```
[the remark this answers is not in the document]
kayra1: done

internal/metrics/metrics.go
kayra1: I am not sure if it is a bug

kayra1: actually we already test this in the metrics test and it works. I'm removing it.

internal/server/handlers_accounts.go
```diff
@@ -165,11 +157,9 @@ func CreateAccount(env *HandlerConfig) http.HandlerFunc {
 			writeError(w, http.StatusInternalServerError, "Internal Error")
 			return
 		}
-		accountResponse := CreateAccountResponse{
-			ID: int(id),
-		}
+		successResponse := SuccessResponse{Message: "success"}
 		w.WriteHeader(http.StatusCreated)
-		err = writeJSON(w, accountResponse)
+		err = writeJSON(w, successResponse)
```
gruyaume, line 162: Shouldn't Notary include the account ID in the Create Account response?

internal/server/handlers_certificate_requests.go
```diff
@@ -102,11 +80,9 @@ func CreateCertificateRequest(env *HandlerConfig) http.HandlerFunc {
 			writeError(w, http.StatusInternalServerError, "Internal Error")
 			return
 		}
-		certificateRequestResponse := CreateCertificateRequestResponse{
-			ID: int(id),
-		}
+		successResponse := SuccessResponse{Message: "success"}
 		w.WriteHeader(http.StatusCreated)
-		err = writeJSON(w, certificateRequestResponse)
+		err = writeJSON(w, successResponse)
```
gruyaume, line 85: I'm not necessarily against all of those HTTP API changes but why include those in this PR, they are quite unrelated to the introduction of sqlair

gruyaume (changes requested):
The sqlair changes look pretty good! Why include Notary HTTP API changes in this PR though? I don't necessarily disagree with those but they don't match the PR description and unnecessarily complexify the PR, and the review process.

internal/server/handlers_certificate_requests.go
kayra1: the db sqlair response no longer returned any integer for me to pass along to the http server. My other option was to return a fake 0 for ids for everything instead, which meant that we would have committed a version of notary where the responses for the API were meaningless.

internal/server/handlers_accounts.go
kayra1: that value is no longer returned by sqlair

internal/server/handlers_certificate_requests_test.go
kayra1: oops, gone

internal/server/handlers_certificate_requests.go
gruyaume: Ok yeah no we don't want to return a field that means nothing either. Let's keep it as is

gruyaume (approved):

ui/src/app/(notary)/certificate_requests/table.tsx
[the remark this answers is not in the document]
saltiyazan: Yes, I suggested the wrong thing by mistake.
I wanted to suggest saying `Revoke Certificate` instead of `Revoke Certificate Request`

saltiyazan (approved):
Looks good, given [this](https://github.com/canonical/notary/pull/122#discussion_r1849653808) comment gets resolved.

PR #132, opened 2024-12-11
feat: certificates table migration

kayra1:
# Description

This change reworks the database to utilize a new certificates table instead of embedding the certificate chain into the CSR table. This new way of doing things will not only allow notary to store and manage orphaned certificates that may have been renewed, but also provide a nice unified way of storing issuing certificates and helping facilitate their signing process.

# Checklist:


internal/db/db_test.go
```diff
@@ -206,13 +30,13 @@ func Example() {
...
-	if entry.CertificateChain != BananaCert {
-		log.Fatalln("Retrieved Certificate doesn't match Stored Certificate")
-	}
+	// entry, err := db.GetCertificateRequestByCSR(BananaCSR)
+	// if err != nil {
+	// 	log.Fatalln(err)
+	// }
+	// if entry.CertificateChain != BananaCert {
+	// 	log.Fatalln("Retrieved Certificate doesn't match Stored Certificate")
+	// }
```
ghislainbourgeois, line 39: Can we also remove this?

ui/package-lock.json
```diff
@@ -7905,6 +7905,21 @@
...
+      "cpu": [
+        "ia32"
+      ],
+      "optional": true,
+      "os": [
+        "win32"
+      ],
+      "engines": {
+        "node": ">= 10"
+      }
```
ghislainbourgeois, line 7922: Is there any reason js dependencies are updated? I don't see any changes to the js parts of the project.

internal/db/db_users_test.go
```diff
@@ -0,0 +1,72 @@
+package db_test
+
+import (
+	"testing"
+
+	"github.com/canonical/notary/internal/db"
+	"golang.org/x/crypto/bcrypt"
+)
+
+func TestUsersEndToEnd(t *testing.T) {
```
ghislainbourgeois, line 10: While I think those tests have just been refactored to their own files, I think they only cover the happy path and do not test the error cases. I would be fine with creating a separate task to fix this in a separate PR.

internal/db/db_certificates_test.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		t.Fatalf("Couldn't complete Update to revoke certificate: %s", err)
+	}
+	retrievedCSR, _ = db.GetCertificateRequestByCSR(BananaCSR)
+	if retrievedCSR.Status != "Revoked" {
+		t.Fatalf("Couldn't revoke certificate")
+	}
+}
+
+func TestUpdateFails(t *testing.T) {
```
ghislainbourgeois, line 83: I think we are lacking more failure scenarios test cases.

ghislainbourgeois (changes requested):

ui/package-lock.json
kayra1: no, I didn't notice this. it looks like it installed a 32 bit version of nextjs's compiler. is it ok to just keep it?

internal/db/db_test.go
kayra1: gone

internal/db/db_users_test.go
kayra1: Since the users aren't relevant to this PR, i'll create a new ticket for them but i'll add more tests for the certificates.

internal/db/db_certificates_test.go
kayra1: I added a bit more tests, but it's still not comprehensive because Notary doesn't actually have certificate management capabilities. I am planning on writing a bunch more tests during the notary as a CA task where certificates will be able to be directly managed.

ui/package-lock.json
ghislainbourgeois: I don't mind, I was just curious.

internal/db/db_certificates.go
```diff
@@ -0,0 +1,293 @@
...
+	}
+	err = db.conn.Query(context.Background(), stmt, certRow).Get(&certRow)
+	if err != nil {
+		return nil, err
+	}
+	return &certRow, nil
+}
+
+// AddCertificateChainToCertificateRequestByCSR adds a new certificate chain to a row for a given CSR string.
+func (db *Database) AddCertificateChainToCertificateRequestByCSR(csrPEM string, certPEM string) error {
```
ghislainbourgeois, line 110: This function and the next one do almost completely the same thing, could we reuse most of the code by extracting a function and passing the CSR after looking it up either by ID or CSR?

ghislainbourgeois (changes requested):

internal/db/db_certificates.go
kayra1: I did something a bit different, please take a look.

ghislainbourgeois (approved):

PR #150, opened 2025-01-23
feat: add CA functionality to the API

kayra1:
# Description

Please include a summary of the change. Please also include relevant motivation and context. List any dependencies that are required for this change.

# Checklist:


internal/db/db_certificate_authorities.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+	return &CARow, nil
+}
+
+// GetCertificateByID gets a certificate row from the repository from a given ID.
+func (db *Database) GetDenormalizedCertificateAuthority(filter CertificateAuthorityFilter) (*CertificateAuthorityDenormalized, error) {
+	var CADenormalizedRow CertificateAuthorityDenormalized
+	var CARow CertificateAuthority
+
+	switch {
```
gruyaume, line 144: (same comment on switch case)

internal/db/db_certificate_authorities.go
```diff
@@ -0,0 +1,276 @@
...
+		return nil, err
+	}
+	return CAs, nil
+}
+
+// GetCertificateByID gets a certificate row from the repository from a given ID.
+func (db *Database) GetCertificateAuthority(filter CertificateAuthorityFilter) (*CertificateAuthority, error) {
+	var CARow CertificateAuthority
+
+	switch {
```
gruyaume, line 117: Is it really a good use of switch cases? This is mostly a personal opinion but switch cases make more sense when we are looking at different options for a given value.

internal/db/db_certificate_authorities.go
```diff
@@ -0,0 +1,276 @@
...
+	}
+	err = db.conn.Query(context.Background(), stmt, CARow, CADenormalizedRow).Get(&CADenormalizedRow)
+	if err != nil {
+		return nil, err
+	}
+	return &CADenormalizedRow, nil
+}
+
+func (db *Database) CreateCertificateAuthority(csrPEM string, privPEM string, certChainPEM string) error {
+	// TODO: It is possible to sign this CSR outside of the Certificate Authority flow.
```
gruyaume, line 166: I'm not sure I understand this TODO, the statement doesn't state what is to be done.

internal/db/db_certificate_authorities.go
```diff
@@ -0,0 +1,276 @@
...
+	ca.Status = status
+	stmt, err := sqlair.Prepare(fmt.Sprintf(updateCertificateAuthorityStmt, db.certificateAuthoritiesTable), CertificateAuthority{})
+	if err != nil {
+		return err
+	}
+	err = db.conn.Query(context.Background(), stmt, ca).Run()
+	return err
+}
+
+// DeleteCertificate removes a certificate from the database.
```
gruyaume, line 255: Not just the certificate right?

internal/db/db_csrs.go
```diff
@@ -254,6 +254,9 @@ func (db *Database) RevokeCertificate(filter CSRFilter) error {
 		CertificateID: 0,
 		Status:        "Revoked",
 	}
+
+	// TODO: CRL stuff
```
gruyaume, line 258: Not the best TODO

internal/db/db_private_keys.go
```diff
@@ -0,0 +1,111 @@
...
+
+const (
+	listPrivateKeysStmt  = "SELECT &PrivateKey.* FROM %s"
+	getPrivateKeyStmt    = "SELECT &PrivateKey.* FROM %s WHERE private_key_id==$PrivateKey.private_key_id or private_key==$PrivateKey.private_key"
+	createPrivateKeyStmt = "INSERT INTO %s (private_key) VALUES ($PrivateKey.private_key)"
+	deletePrivateKeyStmt = "DELETE FROM %s WHERE private_key_id==$PrivateKey.private_key_id or private_key==$PrivateKey.private_key"
+)
+
+// ListCertificateRequests gets every CertificateRequest entry in the table.
+func (db *Database) ListPrivateKeys() ([]PrivateKey, error) {
```
gruyaume, line 33: Is this used anywhere?

internal/db/db_private_keys.go
```diff
@@ -0,0 +1,111 @@
...
+	if err != nil {
+		if errors.Is(err, sqlair.ErrNoRows) {
+			return privateKeys, nil
+		}
+		return nil, err
+	}
+	return privateKeys, nil
+}
+
+// GetCertificateByID gets a certificate row from the repository from a given ID.
```
gruyaume, line 49: Bad description

internal/db/db_certificate_authorities.go
```diff
@@ -0,0 +1,276 @@
...
+	err = db.CreatePrivateKey(privPEM)
+	if err != nil {
+		return err
+	}
+	if certChainPEM != "" {
+		err = db.AddCertificateChainToCertificateRequest(ByCSRPEM(csrPEM), certChainPEM)
+		if err != nil {
+			return err
+		}
+	}
```
gruyaume, line 180: I'm not sure about this `if`. It seems like we could have 2 different functions with independent purposes.

Ex. `createCertificateAuthorityRequest` and `addCertificateToCertificateAuthority` .

Those 2 actions may or may bot be called one after the other (depending on self-signed or not), and I see them as 2 independent things.

internal/server/router.go
```diff
@@ -15,8 +15,11 @@ func NewHandler(config *HandlerConfig) http.Handler {
 	apiV1Router.HandleFunc("POST /certificate_requests", adminOrUser(config.JWTSecret, CreateCertificateRequest(config)))
 	apiV1Router.HandleFunc("GET /certificate_requests/{id}", adminOrUser(config.JWTSecret, GetCertificateRequest(config)))
 	apiV1Router.HandleFunc("DELETE /certificate_requests/{id}", adminOrUser(config.JWTSecret, DeleteCertificateRequest(config)))
+	// apiV1Router.HandleFunc("POST /certificate_requests/{id}/sign", adminOrUser(config.JWTSecret, SignCertificateRequest(config))) TODO
```
gruyaume, line 18: Why those commented out routers? Please avoid those

internal/server/router.go
```diff
@@ -15,8 +15,11 @@ func NewHandler(config *HandlerConfig) http.Handler {
...
+	apiV1Router.HandleFunc("POST /certificate_requests/{id}/reject", adminOrUser(config.JWTSecret, RejectCertificateRequest(config)))
 	apiV1Router.HandleFunc("POST /certificate_requests/{id}/certificate", adminOrUser(config.JWTSecret, CreateCertificate(config)))
-	apiV1Router.HandleFunc("POST /certificate_requests/{id}/certificate/reject", adminOrUser(config.JWTSecret, RejectCertificate(config)))
+	apiV1Router.HandleFunc("POST /certificate_requests/{id}/certificate/renew", adminOrUser(config.JWTSecret, RenewCertificate(config)))
+	// apiV1Router.HandleFunc("POST /certificate_requests/{id}/certificate/revoke", adminOrUser(config.JWTSecret, RevokeCertificate(config))) TODO
```
gruyaume, line 22: Same comment - please avoid those commented out blocks of code

internal/server/router.go
```diff
@@ -25,6 +28,15 @@ func NewHandler(config *HandlerConfig) http.Handler {
 	apiV1Router.HandleFunc("DELETE /accounts/{id}", adminOnly(config.JWTSecret, DeleteAccount(config)))
 	apiV1Router.HandleFunc("POST /accounts/{id}/change_password", adminOrMe(config.JWTSecret, ChangeAccountPassword(config)))
 
+	apiV1Router.HandleFunc("GET /certificate_authorities", adminOnly(config.JWTSecret, ListCertificateAuthorities(config)))
+	apiV1Router.HandleFunc("POST /certificate_authorities", adminOnly(config.JWTSecret, CreateCertificateAuthority(config)))
+	apiV1Router.HandleFunc("GET /certificate_authorities/{id}", adminOnly(config.JWTSecret, GetCertificateAuthority(config)))
+	apiV1Router.HandleFunc("PUT /certificate_authorities/{id}", adminOnly(config.JWTSecret, UpdateCertificateAuthority(config)))
+	apiV1Router.HandleFunc("DELETE /certificate_authorities/{id}", adminOnly(config.JWTSecret, DeleteCertificateAuthority(config)))
+	// apiV1Router.HandleFunc("POST /certificate_authorities/{id}/renew", adminOnly(config.JWTSecret, RenewCertificateAuthority(config))) TODO
```
gruyaume, line 36: Please avoid commented out code blocks

internal/server/handlers_certificate_requests.go
```diff
@@ -265,3 +265,48 @@ func DeleteCertificate(env *HandlerConfig) http.HandlerFunc {
...
+		successResponse := SuccessResponse{Message: "success"}
+		err = writeResponse(w, successResponse, http.StatusAccepted)
+		if err != nil {
+			writeError(w, http.StatusInternalServerError, "internal error")
+			return
+		}
+	}
+}
+
+func SignCertificateRequest(env *HandlerConfig) http.HandlerFunc {
```
gruyaume, line 302: There are many TODO's all over the place, can we remove those.

internal/server/handlers_certificate_authorities.go
```diff
@@ -0,0 +1,342 @@
...
+			writeError(w, http.StatusInternalServerError, "Internal Error")
+			return
+		}
+
+		var params UpdateCertificateAuthorityParams
+		if err := json.NewDecoder(r.Body).Decode(&params); err != nil {
+			writeError(w, http.StatusBadRequest, "Invalid JSON format")
+			return
+		}
+		err = env.DB.UpdateCertificateAuthorityStatus(db.ByCertificateAuthorityID(idNum), params.Status)
```
gruyaume, line 257: We should probably do some input validation before sending it to the db

internal/server/handlers_certificate_authorities.go
```diff
@@ -0,0 +1,342 @@
...
+		}
+		err = writeResponse(w, SuccessResponse{Message: "success"}, http.StatusOK)
+		if err != nil {
+			writeError(w, http.StatusInternalServerError, "internal error")
+			return
+		}
+	}
+}
+
+func RenewCertificateAuthority(env *HandlerConfig) http.HandlerFunc {
```
gruyaume, line 332: Please avoid those functions - they don't do anything

gruyaume (changes requested):

internal/db/db_private_keys.go
```diff
@@ -0,0 +1,111 @@
...
+		return nil, err
+	}
+	err = db.conn.Query(context.Background(), stmt, pkRow).Get(&pkRow)
+	if err != nil {
+		return nil, err
+	}
+	return &pkRow, nil
+}
+
+// CreateCertificateRequest creates a new CSR entry in the repository. The string must be a valid CSR and unique.
```
ghislainbourgeois, line 73: This does not match the function.

internal/db/db_private_keys.go
```diff
@@ -0,0 +1,111 @@
...
+		return err
+	}
+	row := PrivateKey{
+		PrivateKeyPEM: pk,
+	}
+	err = db.conn.Query(context.Background(), stmt, row).Run()
+	return err
+}
+
+// CreateCertificateRequest creates a new CSR entry in the repository. The string must be a valid CSR and unique.
```
ghislainbourgeois, line 89: This does not match the function.

internal/db/validation.go
```diff
@@ -102,6 +102,28 @@ func CertificateMatchesCSR(cert string, csr string) error {
...
+	if block.Type != "RSA PRIVATE KEY" && block.Type != "PRIVATE KEY" {
+		return fmt.Errorf("invalid PEM block type: %s", block.Type)
+	}
+
+	var err error
+	if block.Type == "RSA PRIVATE KEY" {
+		_, err = x509.ParsePKCS1PrivateKey(block.Bytes)
+	} else {
+		_, err = x509.ParsePKCS8PrivateKey(block.Bytes)
+	}
```
ghislainbourgeois, line 120: Could we simplify this code by first trying to parse PKCS8, then try PKCS1 on error? When reading the documentation, it does not seem 100% certain that the PEM block will open with those strings.

internal/server/handlers_certificate_authorities.go
```diff
@@ -0,0 +1,342 @@
...
+
+type UpdateCertificateAuthorityParams struct {
+	Status string `json:"status,omitempty"`
+}
+
+type UploadCertificateToCertificateAuthorityParams struct {
+	CertificateChain string `json:"certificate_chain"`
+}
+
+// createCACertificate uses the input fields from the CA certificate generation form to create
```
ghislainbourgeois, line 51: The name here does not match the function.

internal/server/handlers_certificate_authorities.go
```diff
@@ -0,0 +1,342 @@
...
+	// Create the private key for the CA
+	priv, err := rsa.GenerateKey(rand.Reader, 4096)
+	if err != nil {
+		return "", "", ""
+	}
+
+	privPEM := new(bytes.Buffer)
+	err = pem.Encode(privPEM, &pem.Block{
+		Type:  "RSA PRIVATE KEY",
+		Bytes: x509.MarshalPKCS1PrivateKey(priv),
```
ghislainbourgeois, line 63: I am not sure that's what we should be using, I think PKCS8 is more modern and secure. IANAC (I am not a cryptographer)

internal/server/handlers_certificate_authorities.go
```diff
@@ -0,0 +1,342 @@
...
+		for i, ca := range cas {
+			caResponse[i] = CertificateAuthority{
+				CertificateAuthorityID: ca.CertificateAuthorityID,
+				Status:                 ca.Status,
+				PrivateKeyPEM:          ca.PrivateKeyPEM,
+				CSRPEM:                 ca.CSRPEM,
+				CertificatePEM:         ca.CertificatePEM,
+			}
+		}
+		err = writeResponse(w, caResponse, http.StatusOK)
```
ghislainbourgeois, line 167: Are we returning private keys in the HTTP response?

internal/server/handlers_certificate_authorities.go
```diff
@@ -0,0 +1,342 @@
...
+		}
+		caResponse := CertificateAuthority{
+			CertificateAuthorityID: ca.CertificateAuthorityID,
+			Status:                 ca.Status,
+			CSRPEM:                 ca.CSRPEM,
+			CertificatePEM:         ca.CertificatePEM,
+			PrivateKeyPEM:          ca.PrivateKeyPEM,
+		}
+
+		err = writeResponse(w, caResponse, http.StatusOK)
```
ghislainbourgeois, line 234: We seem to be returning private keys in the HTTP response here also.

ghislainbourgeois (changes requested):
Here is my first pass review.

README.md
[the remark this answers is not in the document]
kayra1: I have added them to the readme and added the context for the PR in the description.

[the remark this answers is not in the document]
kayra1: done

internal/db/db_certificate_authorities.go
kayra1: The switch cases were accepted in a previous PR where we updated how we did filters for db queries. We used to have 2 functions for each query: for example, `GetXByID(id int)` and `GetXByCSR(csr string)`.The solution to merging these was to create a filter object where we use either the ID field or the CSR field of the object depending on where and what we are doing. We *are* looking at different options for a given filter, and that's why I used a switch here. I do want to refactor this, but I did not think that this PR is not the place to change this paradigm here.

internal/server/handlers_certificate_authorities.go
kayra1: the string is enumerated at the table definition level, so any wrong status will be rejected by the db automatically already

internal/db/db_certificate_authorities.go
kayra1: It means that the CSR is visible in the regular CSR view, so it could be signed using the `/api/v1/certificate_requests/{id}/certificate` endpoint, which would fail to update the CA entry. This bug will be fixed in the signing PR. I have removed the comment.

internal/db/db_private_keys.go
kayra1: Only in the tests

internal/server/router.go
kayra1: gone

internal/server/router.go
kayra1: gone

internal/server/handlers_certificate_authorities.go
kayra1: Gone

internal/server/handlers_certificate_requests.go
kayra1: I do not understand why we don't want todo's in the codebase. They are helpful to me personally because I know exactly where I should pick back up when the change is out of scope of the PR and I know for a fact that I will need to do it and there is a related issue in JIRA.

internal/server/router.go
kayra1: Removed.

internal/db/db_csrs.go
kayra1: gone

internal/db/db_certificate_authorities.go
kayra1: fixed

internal/db/db_certificate_authorities.go
kayra1: we already have a function called UpdateCertificateAuthorityCertificate that I can use right after and get rid of this if statement, but it would add 3 more queries to the already large number of queries that need to be done for this single task, so I avoided using that. There would also be a bit of a downtime between the two where the CA exists but the certificate isn't ready yet where the frontend may show that the CA is pending. This way, the CA is created with the certificate ID in place so there is no time where a self signed CA is pending.

I can still update it to be that way if we want.

internal/db/db_certificate_authorities.go
```diff
@@ -0,0 +1,276 @@
...
+	err = db.AddCertificateChainToCertificateRequest(ByCSRID(ca.CSRID), certChainPEM)
+	if err != nil {
+		return err
+	}
+	insertedCert, err := sanitizeCertificateBundle(certChainPEM)
+	if err != nil {
+		return err
+	}
+	newCert, err := db.GetCertificate(ByCertificatePEM(insertedCert[0]))
```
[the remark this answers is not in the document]
kayra1: actually, line 235 will error since entering empty certificates are not allowed, so this sanitize call will never return a chain less than 2 long. Should I put a comment about this?

internal/db/db_certificates.go
```diff
@@ -117,36 +108,70 @@ func (db *Database) AddCertificateChainToCertificateRequest(csrFilter CSRFilter,
 		return errors.New("cert validation failed: " + err.Error())
 	}
 	parentID := 0
-	for i := len(certBundle) - 1; i >= 0; i-- {
+	// If the certificate is self-signed, create a single certificate that references itself
+	if len(certBundle) == 2 && certBundle[0] == certBundle[1] {
```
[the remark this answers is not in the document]
kayra1: done, and removed the comment above

internal/db/db_certificates.go
```diff
@@ -117,36 +108,70 @@ func (db *Database) AddCertificateChainToCertificateRequest(csrFilter CSRFilter,
...
+			return err
+		}
+		err = db.conn.Query(context.Background(), stmt, certRow).Run()
+		if err != nil {
+			return err
+		}
+		parentID = int(childID)
+	} else {
+		// Otherwise, go through the certificate chain in reverse and add certs as their parents
+		for i := len(certBundle) - 1; i >= 0; i-- {
```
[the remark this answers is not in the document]
kayra1: this is nice, but our gomod file currently specifies 1.22.1 which doesn't have it and i dont think i should touch that in this pr. I can open a ticket instead?

internal/db/db_private_keys.go
kayra1: changed

internal/db/validation.go
kayra1: We can get rid of pkcs1 directly, since we are the only ones making keys for the foreseeable future.

The difference between the two is pkcs1 only accepts rsa keys, and pkcs8 accepts many different types. the issue is pkcs8 has BEGIN PRIVATE KEY at the top, which may not be backwards compatible since pkcs1 keys always start with RSA PRIVATE KEY. That's why I try 1 first, and we generate a 1 here. RSA is still perfectly secure, so PKCS1 is also still secure and more compatible. If you want, we can just do PKCS8 only too.

It does not matter which one we do first. they are mutually exclusive.

internal/server/handlers_certificate_authorities.go
kayra1: fixed

internal/server/handlers_certificate_authorities.go
kayra1: see above

internal/server/handlers_certificate_authorities.go
kayra1: fixed

internal/server/handlers_certificate_authorities.go
kayra1: fixed

internal/db/db_certificates.go
```diff
@@ -117,36 +108,69 @@ func (db *Database) AddCertificateChainToCertificateRequest(csrFilter CSRFilter,
 		return errors.New("cert validation failed: " + err.Error())
 	}
 	parentID := 0
-	for i := len(certBundle) - 1; i >= 0; i-- {
+	if isSelfSigned(certBundle) {
```
gruyaume, line 111: Can we separate the code in AddCertificateChainToCertificateRequest in smaller functions? it seems the content inside each if could be a standalone function - feel free to ignore if you disagree.

internal/db/db_certificates.go
```diff
@@ -117,36 +108,69 @@ func (db *Database) AddCertificateChainToCertificateRequest(csrFilter CSRFilter,
...
-			var outcome sqlair.Outcome
-			err = db.conn.Query(context.Background(), stmt, certRow).Get(&outcome)
+			stmt, err := sqlair.Prepare(fmt.Sprintf(getCertificateStmt, db.certificatesTable), Certificate{})
 			if err != nil {
 				return err
 			}
-			childID, err = outcome.Result().LastInsertId()
-			if err != nil {
+			err = db.conn.Query(context.Background(), stmt, certRow).Get(&certRow)
+			childID := int64(certRow.CertificateID)
```
gruyaume, line 154: Why the type conversions? If the underlying type is int64, I would encourage using it all over

internal/db/db_private_keys.go
gruyaume: We should probably delete it then

internal/db/db_private_keys.go
```diff
@@ -0,0 +1,111 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return &pkRow, nil
+}
+
+// CreatePrivateKey creates a new private key entry in the repository. The string must be a valid private key and unique.
+func (db *Database) CreatePrivateKey(pk string) error {
+	if err := ValidatePrivateKey(pk); err != nil {
+		return errors.New("pk validation failed: " + err.Error())
```
gruyaume, line 76:
```suggestion
		return errors.New("private key validation failed: " + err.Error())
```

internal/db/db_private_keys.go
```diff
@@ -0,0 +1,111 @@
...
+		if errors.Is(err, sqlair.ErrNoRows) {
+			return privateKeys, nil
+		}
+		return nil, err
+	}
+	return privateKeys, nil
+}
+
+// GetPrivateKey gets a private key row from the repository from a given ID or PEM.
+func (db *Database) GetPrivateKey(filter PrivateKeyFilter) (*PrivateKey, error) {
```
gruyaume, line 50: Same comments around filters - I'm not sold on their value, I'd encourage to keep things simple

gruyaume (changes requested):

README.md
```diff
@@ -99,3 +105,19 @@ Notary does not support insecure http connections.
 | `/login`                                               | POST        | Login to the Notary UI                         | username, password |
 | `/status`                                              | GET         | Get the status of the Notary service           |                    |
 | `/metrics`                                             | Get         | Get Prometheus metrics                         |                    |
+
+An example for a CAForm:
+```json
+{
+    "self_signed": true,
+    "common_name": "Example Name",
```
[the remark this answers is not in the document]
ghislainbourgeois: Not really, TLS certificates can be issued for entities that are not domain names. S/MIME is a good example of that, where the common name could be "Guillaume Belanger". It's not the first use case we are going for however.

I still think we should give the example with a domain name.

internal/db/db_certificate_authorities.go
```diff
@@ -0,0 +1,280 @@
...
+		if errors.Is(err, sqlair.ErrNoRows) {
+			return CAs, nil
+		}
+		return nil, err
+	}
+	return CAs, nil
+}
+
+// GetCertificateAuthority gets a certificate authority row from the database.
+func (db *Database) GetCertificateAuthority(filter CertificateAuthorityFilter) (*CertificateAuthority, error) {
```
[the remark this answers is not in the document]
ghislainbourgeois: I think the problem is not really with the filters, but with the conditional. I would propose instead to attach a function to `CertificateAuthorityFilter` to convert it to a template `CertificateAuthority`. We could then replace the whole switch with:

```
CARow, err := filter.AsCertificateAuthority()
if err != nil {
    return nil, fmt.Errorf("Cauliflowers are nasty")
}
```

This removes would remove both the concern about the switch, and also make this function do one thing, and not necessitate duplication of a similar function for all the different fields we might want to use for the search.

internal/db/db_certificates.go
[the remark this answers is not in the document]
ghislainbourgeois: I am fine with opening a ticket and doing this in a separate PR.

internal/db/validation.go
ghislainbourgeois: I am fine in supporting only one to start. As I said, I am not a cryptographer, and for now will support the choice you make. I think this should be properly audited by experts later on anyway.

README.md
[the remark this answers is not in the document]
kayra1: RFC 5280 states that a common name is a string of 1 to 64 characters, and it's only used as a fallback domain name if the dns SAN extension is not defined . There are a bunch of charms that use it in this way. I have updated the test to have a domain name.

internal/db/db_certificates.go
kayra1: this bit of awkwardness stems from my lack of knowledge on how the database connectors work in Go when I first created notary. The response from the database is always a string, but it is convention in go's db connectors to cast it into int64 before it is surfaced. 

I will need to replace all the structs to reflect that, which will cascade into changes into almost every file. I'm planning to do this in a later PR: TLSENG-604

internal/db/db_private_keys.go
kayra1: Should I also delete all of the private key tests? We don't ever get or list the private keys and aren't planning on doing that for a while. I can also move these functions into the tests directly.

internal/db/db_certificate_authorities.go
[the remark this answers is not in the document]
kayra1: >  I would encourage something towards GetCertificateAuthorityFromId, GetCertificateAuthorityFrom<xyz> instead of those filter logics.

This was actually the original version before I changed it to this due to comments in another PR. 

> I would propose instead to attach a function to CertificateAuthorityFilter to convert it to a template CertificateAuthority. We could then replace the whole switch with

This was another thing I wanted to rework before doing the rest of the changes. I wanted to do them in the same pr later, but instead I can introduce this here and then refactor the rest of the code later. I updated it here, and there already is a ticket.

internal/db/db_certificates.go
kayra1: My approach to this is that I don't usually like to split functions into smaller ones unless they are self contained in isolation. 

Here for example, I can split this function into `validateCSRCertificatePair`, `createSelfSignedCertificate`, `createCertificate`, `assignCertificateToCSR`. I want to discourage other developers to come in and create a new certificate without a CSR because that has side effects and implications around Notary's functionality that are not straightforward. If we call `AddCertificateChainToCertificateRequest` in any line in any file in Notary, I can guarantee that it will either move the database to another valid state or give an error message that is completely true. I cannot make that same guarantee for the smaller functions that this would split into. For example, if a developer were to call createCertificate alone, there is a chance that we don't update the CSR or the CA to point to it, and move Notary into an invalid state where we have a "dangling" certificate.

The goal is to make writing wrong code harder.

internal/db/db_private_keys.go
kayra1: See discussion above

internal/db/db_certificates.go
gruyaume: As far as I see it's a small change that affects 4 files:
- https://github.com/canonical/notary/pull/165

internal/db/db_private_keys.go
gruyaume: If those tests don't validate a behavior that's handled by our source code - yes I would delete them.

internal/db/db_certificate_authorities.go
```diff
@@ -0,0 +1,276 @@
...
+	if err != nil {
+		return err
+	}
+	var CARow CertificateAuthority
+	if csr.CertificateID != 0 {
+		CARow = CertificateAuthority{
+			CSRID:         csr.CSR_ID,
+			CertificateID: csr.CertificateID,
+			PrivateKeyID:  pk.PrivateKeyID,
+			Status:        "Active",
```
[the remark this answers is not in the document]
kayra1: Go doesn't really have enums, but I can make consts and we can pinky promise not to use strings :D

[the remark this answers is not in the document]
ghislainbourgeois: There is no enum keyword, but constants and type definition can be used to do the same thing: https://www.sohamkamani.com/golang/enums/#enum-types

ghislainbourgeois (commented):
This looks good to me. I am not explicitly approving because of current unresolved discussions.

internal/db/db_certificate_authorities.go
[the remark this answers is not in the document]
ghislainbourgeois: I see that you already did that, closing this conversation.

ghislainbourgeois (approved):

internal/db/db_private_keys.go
kayra1: They do validate that we have successfully created a private key, which we need to do. We just don't have to query it yet.

internal/db/db_certificates.go
kayra1: There's a bit more than that to cover all of the ID fields in the codebase, which should all be int64. I've pushed a commit to that PR that handles those too. It affected 12 files. I can take those changes here, or we can merge that PR after this one, which was what I was planning on doing anyway.

gruyaume (approved):

internal/db/db_certificates.go
gruyaume: Ok I'm fine with the other ones being in a different PR. I mostly wanted to avoid any unnecessary type conversions that we were introducing here.

PR #156, opened 2025-02-07
feat: block large requests from the API

### LINKED ISSUE #117: Validate input size
### Enhancement Proposal

We should validate the input size for certificates, csrs, and very very likely all inputs, and ensure they have a min, max size. This is to ensure attackers don't have the ability to increase the memory usage of notary. For instance, we know exactly what's a reasonable csr size, we can return quickly if the size does not match expectations.

kayra1:
# Description

This change introduces a new middleware that intercepts large requests and rejects them according to the content-size header. Fixes #117.

This was an interesting exploration. There are 2 main options:
* Check the `Content-Length` header
This method is reliable most of the time because this value is autogenerated by all HTTP clients. However, it is possible to modify this header by a malicious actor.

* Read the body and measure the length.
 There is already a 10 MB limit to all request bodies in Go. Editing this limit requires intercepting the reader buffer with a `MaxBytesReader`.This wraps a buffer with another buffer, that internally counts the amount of data read and closes the buffer and returns an error if the custom limit is reached. Since this error only materializes after reading the buffer (which erases the data in it), we need to read the whole buffer until we reach our maxbytes or the buffer ends, and then place the extracted value into another buffer for the actual HTTP handler to process the request.

This method is the most reliable out of all of the options, but requires reading the body twice. I preferred this method in this PR because it is relatively cheap to read a couple kilobytes of data twice, and the likelihood of a malicious actor doing this is very low.

# Checklist:


internal/server/router.go
```diff
@@ -36,6 +36,7 @@ func NewHandler(config *HandlerConfig) http.Handler {
 		jwtSecret: config.JWTSecret,
 	}
 	apiMiddlewareStack := createMiddlewareStack(
+		limitRequestSize(100),
```
ghislainbourgeois, line 39: I would suggest making a constant for the value, something like `MAX_KILOBITS`.

ghislainbourgeois (changes requested):
Love the tests! I left a small suggestion, otherwise LGTM

internal/server/router.go
kayra1: done

ghislainbourgeois (approved):

PR #175, opened 2025-02-14
feat: add API input validation and return appropriate HTTP status code

### LINKED ISSUE #170: Internal error when updating certificate authorities
#### Describe the bug

I was testing the new feature introduced in #150 and noticed an issue with updating the certificate authority.

#### To Reproduce

1. Create a Certificate Authority
2. Update the certificate authority using the PUT method 

#### Expected behavior
A reasonable status (ex. 400) and informative content (ex. "missing `status` field).

Normal use of the API, even with bad inputs, should almost never result in 500 status returns.

#### Logs

Notary Logs
```
2025/02/14 09:10:38 CHECK constraint failed: status IN ('active', 'expired', 'pending', 'legacy')
2025/02/14 09:10:38 Internal Error
2025/02/14 09:10:38 PUT /certificate_authorities/1 500 Internal Server Error
```

Postman logs
```
Warning: Self signed certificate
PUT /api/v1/certificate_authorities/1 HTTP/1.1
Content-Type: application/json
Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpZCI6MSwidXNlcm5hbWUiOiJwaXp6YS5jb20iLCJwZXJtaXNzaW9ucyI6MSwiZXhwIjoxNzM5NTQ1ODExfQ.Ipt1EQa6l2sfocZJfJ_ufNyXQSZZy0y6bytCXkUISn8
User-Agent: PostmanRuntime/7.43.0
Accept: */*
Cache-Control: no-cache
Postman-Token: 254effef-c5d1-42af-ab81-eee52e12eb14
Host: 127.0.0.1:3000
Accept-Encoding: gzip, deflate, br
Connection: keep-alive
Content-Length: 3
 
{
}
 
HTTP/1.1 500 Internal Server Error
Content-Type: application/json
Date: Fri, 14 Feb 2025 14:12:04 GMT
Content-Length: 26
 
{"error":"Internal Error"}
```

#### Environment

- Ubuntu and snap
- Notary: 0.0.3
- Snap revision: 130

### LINKED ISSUE #172: No success response when deleting a certificate authority
#### Describe the bug

When deleting a certificate authority, users would expect to get a success response (200) to confirm the deletion was complete. I noticed this issue in the context of testing #150 .

Relevant code:
- https://github.com/canonical/notary/blob/main/internal/server/handlers_certificate_authorities.go#L288

#### To Reproduce

1. Create a Certificate Authority
2. Delete the Certificate Authority

#### Expected behavior
200 Status

#### Environment

- Notary: 0.0.3
- Snap: 130

gruyaume:
# Description

We shouldn't return a 500 error code when the issue comes from the user input. The 400 Bad Request status code is more appropriate and informative in that context. Here, we add proper HTTP API input validation and return a relevant and informative status code when the input is invalid. For now, the changes have only been made for the CA endpoints. Once this PR is approved and merged, I will apply the same approach for all of our endpoints. 

Fixes #170 
Fixes #172 

## Before and after

Here we have the before and after returns when making a PUT call to change the status of a Certificate Authority. 

**Before**

```
Status: 500 Internal error
Response:
{
    "error": "Internal Error"
}
```

**After**

```
Status: 400 Bad Request
Response:
{
    "error": "Invalid request: invalid status: status must be one of active, expired, pending, legacy"
}
```
# Checklist:

### DIFF
--- a/internal/server/handlers_certificate_authorities.go
+++ b/internal/server/handlers_certificate_authorities.go
@@ -9,6 +9,7 @@ import (
 	"encoding/json"
 	"encoding/pem"
 	"errors"
+	"fmt"
 	"log"
 	"math/big"
 	"net/http"
@@ -46,6 +47,35 @@ type UpdateCertificateAuthorityParams struct {
 
 type UploadCertificateToCertificateAuthorityParams struct {
 	CertificateChain string `json:"certificate_chain"`
+}
+
+func (params *CreateCertificateAuthorityParams) IsValid() (bool, error) {
+	// If a country is provided, it must be exactly two letters (ISO 3166-1 alpha-2).
+	if params.CountryName != "" && len(params.CountryName) != 2 {
+		return false, fmt.Errorf("country_name must be a 2-letter ISO code")
+	}
+
+	// Validate the NotValidAfter field if provided.
+	if params.NotValidAfter != "" {
+		notValidAfter, err := time.Parse(time.RFC3339, params.NotValidAfter)
+		if err != nil {
+			return false, fmt.Errorf("not_valid_after must be a valid RFC3339 timestamp")
+		}
+		if !notValidAfter.After(time.Now()) {
+			return false, errors.New("not_valid_after must be a future time")
+		}
+	}
+	return true, nil
+}
+
+func (updateCAParams *UpdateCertificateAuthorityParams) IsValid() (bool, error) {
+	if updateCAParams.Status == "" {
+		return false, errors.New("status is required")
+	}
+	if updateCAParams.Status != db.CAActive && updateCAParams.Status != db.CAExpired && updateCAParams.Status != db.CAPending && updateCAParams.Status != db.CALegacy {
+		return false, fmt.Errorf("Invalid status. Status must be one of %s, %s, %s, %s", db.CAActive, db.CAExpired, db.CAPending, db.CALegacy)
+	}
+	return true, nil
 }
 
 // createCertificateAuthority uses the input fields from the CA certificate generation form to create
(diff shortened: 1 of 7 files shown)


internal/server/handlers_certificate_authorities.go
```diff
@@ -48,6 +50,65 @@ type UploadCertificateToCertificateAuthorityParams struct {
...
+		}
+	}
+	return true, nil
+}
+
+func (updateCAParams *UpdateCertificateAuthorityParams) IsValid() (bool, error) {
+	if updateCAParams.Status == "" {
+		return false, errors.New("status is required")
+	}
+	if updateCAParams.Status != db.CAActive && updateCAParams.Status != db.CAExpired && updateCAParams.Status != db.CAPending && updateCAParams.Status != db.CALegacy {
```
ghislainbourgeois, line 76: I would propose adding a function attached to create a `Status` type from a string and error out if it is not a valid string. We could use that function here and reuse it anywhere we want to get a `Status`.

Something like:
```
func NewStatusFromString(s string) (Status, error) {
    statuses := map[Status]struct{}{
        CAActive: {},
        CAExpired: {},
        CAPending: {},
        CALegacy: {},
    }

    status := Status(s)
    _, ok := statuses[status]
    if !ok {
        return nil, fmt.Errorf("invalid status: status must be one of %s, %s, %s, %s", CAActive, CAExpired, CAPending, CALegacy)
    }
    return status, nil
}
```

Then, we could do this here:

```
if _, err := NewStatusFromString(updateCAParams.Status); err != nil {
         return false, err
}
return true, nil
```

ghislainbourgeois (changes requested):
I have small suggestion

internal/server/handlers_certificate_authorities.go
gruyaume: Done!

ghislainbourgeois (approved):

PR #181, opened 2025-02-18
fix: infinite certificate loop

### LINKED ISSUE #176: Notary hangs at `ListCertificateRequestWithCertificates`
#### Describe the bug
When playing around with the API, it is possible to bring the system to a point where it hangs completely. Looking at Parca, it's clear that Notary hangs at `ListCertificateRequestWithCertificates`. This db call is recursive which indicates that it is indeed where the issue is.

#### To Reproduce

1. Create Account
2. Login
3. Create Certificate Authority (self-signed)
4. List Certificate Requests

#### Screenshots

Image

kayra1:
# Description

The reason fixes #176 happens is because we have an infinite loop where a certificate refers to itself if it is self signed. We usually store self signed certificates in the database so that they refer to no certificate instead of their own ID in the database in order to avoid this issue, since we can infer that all certificates without parents are self signed.

# Checklist:

### DIFF
--- a/internal/db/db_certificate_authorities_test.go
+++ b/internal/db/db_certificate_authorities_test.go
@@ -451,3 +451,35 @@ func TestCertificateAuthorityFails(t *testing.T) {
 		t.Fatalf("Should have failed to delete certificate authority")
 	}
 }
+
+func TestSelfSignedCertificateList(t *testing.T) {
+	tempDir := t.TempDir()
+	database, err := db.NewDatabase(filepath.Join(tempDir, "db.sqlite3"))
+	if err != nil {
+		t.Fatalf("Couldn't complete NewDatabase: %s", err)
+	}
+	defer database.Close()
+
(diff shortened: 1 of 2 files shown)


gruyaume (commented):
I'm still somewhat concerned about the risk of infinite recursion in the db table. With this change, you are confident that it's not possible to craft a set of API queries that would lead to such a loop?

kayra1:
We are traversing a tree with this algorithm, so if there is a node that connects to itself, it will cause an infinite loop. This is true for both a recursive algorithm with a base case and an iterative algorithm with a while loop, written in either SQL or Go. 

After this PR, I do not think it's possible to add a node that connects to itself, as I've removed the exception code that allows that and the regular code already blocks duplicate certificates. This may change for future updates coming to notary.

The best we can do is to have more tests that target this specific vulnerability to make sure that regression never happens, or have another algorithm that avoids tree traversal (for example, SELECT * and manually build a chain for each CSR) but that would incur a serious performance cost depending on the algorithm.

gruyaume (approved):

PR #183, opened 2025-02-19
feat: return last inserted id for create type functions

kayra1:
# Description

This change reintroduces the last inserted id parameter, but only to create type functions. Previously, I had determined that this value did not update if we did updates or deletes in SQL, which made the functions confusing. Since then I've learned how this value is populated (it always contains the last inserted id, even if an irrelevant SELECT or UPDATE statement was the actual SQL statement that ran), and have decided to reintroduce it to reduce complexity in parts of the code where we'd have to immediately query for rows we had just inserted because we didn't have a way to know the last inserted ID's.

# Checklist:

### DIFF
--- a/internal/server/response.go
+++ b/internal/server/response.go
@@ -11,8 +11,8 @@ type SuccessResponse struct {
 }
 
 type CreateSuccessResponse struct {
-	Message  string `json:"message"`
-	ObjectID int64  `json:"object_id"`
+	Message string `json:"message"`
+	ID      int64  `json:"id"`
 }
 
 // writeResponse is a helper function that writes a JSON response to the http.ResponseWriter
--- a/internal/server/response.go
+++ b/internal/server/response.go
@@ -10,6 +10,11 @@ type SuccessResponse struct {
 	Message string `json:"message"`
 }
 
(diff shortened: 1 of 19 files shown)


gruyaume (changes requested):
These changes affect the API, please adapt the API documentation accordingly.

kayra1:
done i think, let me know if there's other places where this is documented.

internal/server/response.go
```diff
@@ -10,6 +10,11 @@ type SuccessResponse struct {
 	Message string `json:"message"`
 }
 
+type CreateSuccessResponse struct {
+	Message  string `json:"message"`
+	ObjectID int64  `json:"object_id"`
```
gruyaume, line 15: I'd prefer `id` instead of `object_id`. `object` doesn't bring much

gruyaume (changes requested):
Overall this looks good, I only have one comment around using `id` instead of `object_id`.

internal/server/response.go
kayra1: done

gruyaume (approved):

PR #184, opened 2025-02-19
feat: sign csrs from within Notary

kayra1:
# Description

Please include a summary of the change. Please also include relevant motivation and context. List any dependencies that are required for this change.

# Checklist:

### DIFF
--- a/internal/db/db_certificate_authorities.go
+++ b/internal/db/db_certificate_authorities.go
@@ -183,50 +183,46 @@ func (db *Database) GetDenormalizedCertificateAuthority(filter CertificateAuthor
 
 // CreateCertificateAuthority creates a new certificate authority in the database from a given CSR, private key, and certificate chain.
 // The certificate chain is optional and can be empty.
-func (db *Database) CreateCertificateAuthority(csrPEM string, privPEM string, certChainPEM string) error {
-	err := db.CreateCertificateRequest(csrPEM)
+func (db *Database) CreateCertificateAuthority(csrPEM string, privPEM string, certChainPEM string) (int64, error) {
+	csrID, err := db.CreateCertificateRequest(csrPEM)
 	if err != nil {
-		return err
+		return 0, err
 	}
-	err = db.CreatePrivateKey(privPEM)
+	pkID, err := db.CreatePrivateKey(privPEM)
 	if err != nil {
-		return err
+		return 0, err
+	}
+	CARow := CertificateAuthority{
+		CSRID:        csrID,
+		PrivateKeyID: pkID,
+		Status:       CAPending,
 	}
 	if certChainPEM != "" {
-		err = db.AddCertificateChainToCertificateRequest(ByCSRPEM(csrPEM), certChainPEM)
+		certID, err := db.AddCertificateChainToCertificateRequest(ByCSRID(csrID), certChainPEM)
 		if err != nil {
-			return err
+			return 0, err
 		}
-	}
-	csr, err := db.GetCertificateRequest(ByCSRPEM(csrPEM))
-	if err != nil {
-		return err
-	}
-	pk, err := db.GetPrivateKey(ByPrivateKeyPEM(privPEM))
-	if err != nil {
-		return err
-	}
-	var CARow CertificateAuthority
-	if csr.CertificateID != 0 {
 		CARow = CertificateAuthority{
-			CSRID:         csr.CSR_ID,
-			CertificateID: csr.CertificateID,
-			PrivateKeyID:  pk.PrivateKeyID,
+			CSRID:         csrID,
+			CertificateID: certID,
+			PrivateKeyID:  pkID,
 			Status:        CAActive,
-		}
-	} else {
-		CARow = CertificateAuthority{
-			CSRID:        csr.CSR_ID,
-			PrivateKeyID: pk.PrivateKeyID,
-			Status:       CAPending,
 		}
 	}
 	stmt, err := sqlair.Prepare(createCertificateAuthorityStmt, CertificateAuthority{})
 	if err != nil {
-		return err
+		return 0, err
 	}
-	err = db.conn.Query(context.Background(), stmt, CARow).Run()
-	return err
+	var outcome sqlair.Outcome
+	err = db.conn.Query(context.Background(), stmt, CARow).Get(&outcome)
+	if err != nil {
+		return 0, err
+	}
+	insertedRowID, err := outcome.Result().LastInsertId()
+	if err != nil {
+		return 0, err
+	}
+	return insertedRowID, nil
 }
 
 // UpdateCertificateAuthorityCertificate updates the certificate chain associated with a certificate authority.
@@ -235,19 +231,11 @@ func (db *Database) UpdateCertificateAuthorityCertificate(filter CertificateAuth
 	if err != nil {
 		return err
 	}
-	err = db.AddCertificateChainToCertificateRequest(ByCSRID(ca.CSRID), certChainPEM)
+	certID, err := db.AddCertificateChainToCertificateRequest(ByCSRID(ca.CSRID), certChainPEM)
 	if err != nil {
 		return err
 	}
-	insertedCert, err := sanitizeCertificateBundle(certChainPEM)
-	if err != nil {
-		return err
-	}
-	newCert, err := db.GetCertificate(ByCertificatePEM(insertedCert[0]))
-	if err != nil {
-		return err
-	}
-	ca.CertificateID = newCert.CertificateID
+	ca.CertificateID = certID
 	ca.Status = CAActive
 
 	stmt, err := sqlair.Prepare(updateCertificateAuthorityStmt, CertificateAuthority{})
--- a/internal/db/db_certificate_authorities.go
+++ b/internal/db/db_certificate_authorities.go
@@ -1,10 +1,16 @@
 package db
 
 import (
+	"bytes"
 	"context"
+	"crypto/rand"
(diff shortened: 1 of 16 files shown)


internal/server/router.go
```diff
@@ -16,20 +16,24 @@ func NewHandler(config *HandlerConfig) http.Handler {
 	apiV1Router.HandleFunc("GET /certificate_requests/{id}", adminOrUser(config.JWTSecret, GetCertificateRequest(config)))
 	apiV1Router.HandleFunc("DELETE /certificate_requests/{id}", adminOrUser(config.JWTSecret, DeleteCertificateRequest(config)))
 	apiV1Router.HandleFunc("POST /certificate_requests/{id}/reject", adminOrUser(config.JWTSecret, RejectCertificateRequest(config)))
+	apiV1Router.HandleFunc("POST /certificate_requests/{id}/sign", adminOrUser(config.JWTSecret, SignCertificateRequest(config)))
```
gruyaume, line 19: We should update the documentation with this new endpoint.

internal/db/db_certificate_authorities.go
```diff
@@ -274,3 +336,99 @@ func (db *Database) DeleteCertificateAuthority(filter CertificateAuthorityFilter
...
+		certTemplate.ExtKeyUsage = []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth}
+		certTemplate.BasicConstraintsValid = true
+		certTemplate.IsCA = true
+	}
+	certBytes, err := x509.CreateCertificate(rand.Reader, certTemplate, caCertParsed, certTemplate.PublicKey, caPrivateKey)
+	if err != nil {
+		return err
+	}
+	certPEM := new(bytes.Buffer)
+	err = pem.Encode(certPEM, &pem.Block{Type: "CERTIFICATE", Bytes: certBytes})
```
gruyaume, line 418: Ain't there an enum that we can use (from the `pem` library) to specify we want a certificate instead of using a string?

internal/db/db_certificate_authorities.go
```diff
@@ -274,3 +336,99 @@ func (db *Database) DeleteCertificateAuthority(filter CertificateAuthorityFilter
...
+	}
+	block, _ = pem.Decode([]byte(caRow.PrivateKeyPEM))
+	caPrivateKey, err := x509.ParsePKCS1PrivateKey(block.Bytes)
+	if err != nil {
+		return err
+	}
+	if err = certRequest.CheckSignature(); err != nil {
+		return fmt.Errorf("invalid certificate request signature: %w", err)
+	}
+	CSRIsForACertificateAuthority := true
```
gruyaume, line 382: This may not be a big deal but I find this strange for the default value to be that the csr is for a CA, in most cases it won't be.

internal/server/handlers_certificate_authorities.go
```diff
@@ -425,3 +433,98 @@ func PostCertificateAuthorityCertificate(env *HandlerConfig) http.HandlerFunc {
...
+		successResponse := SuccessResponse{Message: "success"}
+		err = writeResponse(w, successResponse, http.StatusAccepted)
+		if err != nil {
+			writeError(w, http.StatusInternalServerError, "internal error")
+			return
+		}
+	}
+}
+
+// SignCertificateAuthority
```
gruyaume, line 482: I'm unconvinced about the value of this comment

gruyaume (changes requested):
Theses are the obvious comments I observed during a quick review. I'll do a proper review, deploying the workload, manually testing ,etc.

internal/server/router.go
kayra1: You gave me 12 minutes :D

internal/db/db_certificate_authorities.go
kayra1: It is only the "default" because it matches the no error state. reversing the statements is produces unexpected code imo:

```
if err != nil {
  if !errors.Is(err, sqlair.ErrNoRows) {
    return err
  }
  CSRIsForACertificateAuthority = false
}
```
becomes
```
if (err == nil) { // this is unexpected
  CSRIsForACertificateAuthority = true
}
if (err != nil && !errors.Is(err, sqlair.ErrNoRows)) {
  return err
}
```

internal/db/db_certificate_authorities.go
```diff
@@ -274,3 +336,99 @@ func (db *Database) DeleteCertificateAuthority(filter CertificateAuthorityFilter
...
+		IPAddresses:        certRequest.IPAddresses,
+		URIs:               certRequest.URIs,
+		DNSNames:           certRequest.DNSNames,
+		PublicKey:          certRequest.PublicKey,
+		PublicKeyAlgorithm: certRequest.PublicKeyAlgorithm,
+
+		// Add standard certificate fields
+		SerialNumber: big.NewInt(time.Now().UnixNano()),
+		NotBefore:    time.Now(),
+		NotAfter:     time.Now().AddDate(1, 0, 0),
```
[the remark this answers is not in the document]
gruyaume: The `1` value can be a const (ex. `expiryYears = 1`)

internal/server/router.go
```diff
@@ -16,20 +16,24 @@ func NewHandler(config *HandlerConfig) http.Handler {
...
-	apiV1Router.HandleFunc("GET /accounts/{id}", adminOrMe(config.JWTSecret, GetAccount(config)))
-	apiV1Router.HandleFunc("DELETE /accounts/{id}", adminOnly(config.JWTSecret, DeleteAccount(config)))
-	apiV1Router.HandleFunc("POST /accounts/{id}/change_password", adminOrMe(config.JWTSecret, ChangeAccountPassword(config)))
+
 	apiV1Router.HandleFunc("GET /certificate_authorities", adminOnly(config.JWTSecret, ListCertificateAuthorities(config)))
 	apiV1Router.HandleFunc("POST /certificate_authorities", adminOnly(config.JWTSecret, CreateCertificateAuthority(config)))
 	apiV1Router.HandleFunc("GET /certificate_authorities/{id}", adminOnly(config.JWTSecret, GetCertificateAuthority(config)))
 	apiV1Router.HandleFunc("PUT /certificate_authorities/{id}", adminOnly(config.JWTSecret, UpdateCertificateAuthority(config)))
 	apiV1Router.HandleFunc("DELETE /certificate_authorities/{id}", adminOnly(config.JWTSecret, DeleteCertificateAuthority(config)))
+	apiV1Router.HandleFunc("POST /certificate_authorities/{id}/sign", adminOnly(config.JWTSecret, SignCertificateAuthority(config)))
```
gruyaume, line 28: I'm unsure about this new endpoint, a Certificate Authority could have multiple certificates over its life right? If that's the case, what does `/sign` sign? I think `/sign` makes sense against a certificate request object, not necessarily against a certificate authority object.

internal/server/handlers_certificate_authorities.go
gruyaume, line 1: In the API, creating a Certificate Authority object has the side effect of creating a Certificate Request object.
```
- POST `api/v1/certificate_authority`
- GET `api/v1/certificate_requests`:
  - You see that Notary returns certificate requests even though you never made a POST against certificate_requests. 
```

I don't think that's the design we should go for. Namely because of these issues: 
- Users can delete a CA CSR by making a DELETE call against the `api/v1/certificate_requests/{id}` for the CA certificate request. What does this intent represent?
- Two possible ways to sign a CA CSR: are by making a POST call to `/certificate_authorities/{id}/certificate` or to `/certificate_requests/{id}/certificate`.
- It's confusing. New API objects are created without user intent.

I recommend that all CA operations (ex., viewing, signing, renewing, etc.) should be done exclusively through the `certificate_authorities` endpoint. 

I understand there's a lot in common between CA Certificate Requests and external certificate requests, but I don't think commonality should be exposed through the API like it is now.

gruyaume (changes requested):

internal/db/db_certificate_authorities.go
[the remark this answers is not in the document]
kayra1: done

internal/server/handlers_certificate_authorities.go
kayra1: I've made it so the endpoint only returns certificate requests, not certificate signing requests. We send the same error as if we could not. 

The error handling got a bit jumbled up until this point, so I've also made TLSENG-643 to fix that later

internal/db/db_certificate_authorities.go
kayra1: I still ended up updating it with helper functions. It is a bit awkward and ugly, so I'm planning on tackling it when working on TLSENG-643

internal/server/handlers_certificate_authorities.go
gruyaume, line 1: I get an internal error when trying to Get a certificate request that was never created

To reproduce:
- POST Certificate Authority
- GET Certificate request with id `1`

## Actual outcome:
```
{
    "error": "Internal Error"
}
```

## Expected outcome

```
{
    "error": "Not found"
}
```

## Discussion

Like we discussed, at the API level the certificate requests objects represent external certificate requests, and shouldn't be intertwined with CA CSR's.

I understand we want to reduce code duplication - but it shouldn't come at the expense of making the system intuitive for users.

gruyaume (changes requested):

internal/server/handlers_certificate_authorities.go
kayra1: Done. I will tackle this further in TLSENG-643 as well.

gruyaume (approved):

PR #194, opened 2025-02-28
feat: Increase bcrypt cost for higher security

ghislainbourgeois:
# Description

Bcrypt's cost is used to tune the password hashing algorithm's compute time. For better protection against brute-force attacks, it should be set higher, at the cost of performance.
A good rule of thumb is to set this so that it takes over 250 ms to hash a password on common hardware.

I added a benchmark test that can be used to evaluate the value. I selected the proposed value in this PR so that each operation takes over 250 ms on my laptop.

The previous value was set by the language to 10, and it took only 43 ms per hash on my laptop.

A future improvement here would be to make this cost configurable. Another improvement would be to automatically increase this cost as time moves on, and rehash passwords with the new cost on logins.

# Checklist:

### DIFF
--- a/internal/db/validation.go
+++ b/internal/db/validation.go
@@ -12,6 +12,13 @@ import (
 	"golang.org/x/crypto/bcrypt"
 )
 
+// BCRYPT_COST is used to tune the password hashing algorithm's time.
+// For better protection against brute-force attacks, this should be
+// set higher, at the cost of performance.
+// A good rule of thumb is to set this so that it takes over 250 ms
+// to hash a password on common hardware.
+const BCRYPT_COST = 13
+
 // ValidateCertificateRequest validates the given CSR string to the following:
 // The string must be a valid PEM string, and should be of type CERTIFICATE REQUEST
 // The PEM string should be able to be parsed into a x509 Certificate Request
@@ -149,7 +156,7 @@ func HashPassword(password string) (string, error) {
 	if strings.TrimSpace(password) == "" {
 		return "", fmt.Errorf("password cannot be empty")
 	}
-	hashedPassword, err := bcrypt.GenerateFromPassword([]byte(password), bcrypt.DefaultCost)
+	hashedPassword, err := bcrypt.GenerateFromPassword([]byte(password), BCRYPT_COST)
 	if err != nil {
 		return "", err
 	}
--- a/internal/db/validation_test.go
+++ b/internal/db/validation_test.go
@@ -418,3 +418,9 @@ func TestCertificateMatchesCSRFail(t *testing.T) {
 		})
 	}
 }
+
+func BenchmarkHashPassword(b *testing.B) {
+	for i := 0; i < b.N; i++ {
+		db.HashPassword("Correct Staple Horse")
+	}
+}
--- a/internal/db/validation_test.go
+++ b/internal/db/validation_test.go
@@ -421,6 +421,6 @@ func TestCertificateMatchesCSRFail(t *testing.T) {
 
 func BenchmarkHashPassword(b *testing.B) {
 	for i := 0; i < b.N; i++ {
-		db.HashPassword("Correct Staple Horse")
+		db.HashPassword("Correct Staple Horse") // nolint:errcheck
 	}
 }


internal/db/validation.go
gruyaume, line 1: Is there any security related concern here where anyone could send tons of requests to our `/login` endpoint and seriously impact our system.

With hundreds of parallel login requests I get the system to use 100% CPU.

That problem was likely already there though, but this PR will for sure make it worse. We should probably keep this in mind for the future and plan for adding ways to prevent this abuse. 

## Screenshots

image

image

gruyaume (commented):
I have no strong comment here, mostly points of discussions related to secuity and performance.

internal/db/validation.go
ghislainbourgeois: Yes, this is a threat that we should take into account, probably by throttling login attempts after failures.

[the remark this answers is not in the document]
ghislainbourgeois: Agreed, we could also change the value when running the tests.

kayra1:
Wanted to give some context about this.

bcrypt's hash cost slows down the hashing algorithm exponentially as the number increases. The reason this exists is if the password ever leaks and the attacker has access to the hashed string, they will attempt to hash every possible 8+ length string to find the matching result.

Our passwords are, at minimum, 8 characters long with one lowercase, one uppercase, and one digit or special character. At 43 ms per hash ,that's 23.2 hashes per second on Ghislain's laptop. 

with some napkin math, a password that contains lowercase, uppercase, digits and special characters has 7.026 x 10^15 variations

it takes Ghislain ~10000 years to crack a password with at least 1 lowercase, 1 uppercase, 1 digit or 1 special character password on default cost.

I looked up how other more powerful hardware perform, and found one that used 8x GTX titan's from 2015. https://security.stackexchange.com/questions/139721/estimate-the-time-to-crack-passwords-using-bcrypt

It would take this setup ~66 years to crack a notary password.

at bcrypt cost 13, this setup would take ~500 years to crack a notary password.

I can run some real benchmarks on my 4090 if we want to actually know the real result. Also, if the performance hit is too much, we can also increase the minimum password length to 10, which would be equivalent to increasing the cost to 13.

internal/db/validation.go
DanielArndt: > When the attacker is using the wrong username it's not as bad.

Is that in itself a security concern? is an adversary able to detect if a user exists by comparing how long it takes us to return on login attempts?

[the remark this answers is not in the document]
DanielArndt: I like the idea of reducing it for tests.

ghislainbourgeois: > > When the attacker is using the wrong username it's not as bad.
> 
> Is that in itself a security concern? is an adversary able to detect if a user exists by comparing how long it takes us to return on login attempts?

I agree that this seems like a potential issue. I will open a ticket for this.

ghislainbourgeois:
> Wanted to give some context about this.
> 
> bcrypt's hash cost slows down the hashing algorithm exponentially as the number increases. The reason this exists is if the password ever leaks and the attacker has access to the hashed string, they will attempt to hash every possible 8+ length string to find the matching result.
> 
> Our passwords are, at minimum, 8 characters long with one lowercase, one uppercase, and one digit or special character. At 43 ms per hash ,that's 23.2 hashes per second on Ghislain's laptop.
> 
> with some napkin math, a password that contains lowercase, uppercase, digits and special characters has 7.026 x 10^15 variations
> 
> it takes Ghislain ~10000 years to crack a password with at least 1 lowercase, 1 uppercase, 1 digit or 1 special character password on default cost.
> 
> I looked up how other more powerful hardware perform, and found one that used 8x GTX titan's from 2015. https://security.stackexchange.com/questions/139721/estimate-the-time-to-crack-passwords-using-bcrypt
> 
> It would take this setup ~66 years to crack a notary password.
> 
> at bcrypt cost 13, this setup would take ~500 years to crack a notary password.
> 
> I can run some real benchmarks on my 4090 if we want to actually know the real hash rate with bcrypt(10) and bcrypt(13) and bcrypt(X) to have a better idea. Also, if the performance hit is too much, we can also increase the minimum password length to 10, which would be equivalent to increasing the cost when it comes to preventing brute force attacks.
> 
> I am skeptical that this is a real increase in security. If the good rule of thumb is 250ms on a users machine, the rule of thumb would be radically different depending on if it's for a framework laptop, the company lenovo laptop, and my desktop with a 4090 GPU.

I tried running hashcat on my corporate laptop running an NVIDIA RTX A500 and I am able to run around 685 H/s with a value of 10. On my desktop with a 3080TI, I can run around 3300 H/s with the same value.

If I try with a value of 13, the laptop's rate drops to 86 H/s and the desktop drops to 438 H/s. That's almost 8 times slower, making the attack a bit more impractical.

Your time estimation also assumes that the passwords are perfectly random following the composition rules. Those rules in fact tend to make passwords less random, not more. People being people, we also need to take into account other things like dictionary attacks. As an extreme example, the password `Qwerty123` satisfies all the composition rules Notary has in place. However, using a word list from the internet containing 4 million passwords, I can crack it in less than 2 seconds.

Looking at the [OWASP cheatsheet](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html), `bcrypt` is recommended only with a work factor of minimum 10. According to it, we should probably revisit this and use Argon2id. If we want to eventually support FIPS, we should switch or add support for PBKDF2.

In conclusion, increasing the work factor of bcrypt does increase security in case an attacker gets access to the hashed passwords, assuming our users are human and do not use purely random passwords. We should probably reconsider bcrypt in the future.

kayra1:
> If I try with a value of 13, the laptop's rate drops to 86 H/s and the desktop drops to 438 H/s. That's almost 8 times slower, making the attack a bit more impractical.

I think 8 times slower is not really a significant enough difference to deter someone from cracking a hash. 1 minute is probably just as acceptable to wait for an attacker as 8 minutes. 1 hour is just as acceptable as 8 hours. 1 day is just as acceptable as 8 days etc.

Also just like you said, `Qwerty123` would be cracked in 16 seconds. To be clear I'm not opposed to merging this or increasing the cost in general.

I was not aware that bcrypt is outdated. I would be happy to move to Argon2id.

ghislainbourgeois:
> > If I try with a value of 13, the laptop's rate drops to 86 H/s and the desktop drops to 438 H/s. That's almost 8 times slower, making the attack a bit more impractical.
> 
> I think 8 times slower is not really a significant enough difference to deter someone from cracking a hash. 1 minute is probably just as acceptable to wait for an attacker as 8 minutes. 1 hour is just as acceptable as 8 hours. 1 day is just as acceptable as 8 days etc.
> 
> Also just like you said, `Qwerty123` would be cracked in 16 seconds. To be clear I'm not opposed to merging this or increasing the cost in general.
> 
> I was not aware that bcrypt is outdated. I would be happy to move to Argon2id.

The increase in time, even though it is not huge, makes the attack less practical and buys time to realize their was a breach and change passwords. Argon2id for example is purposefully slow, but also memory hard to negate GPU attacks.

With my research from yesterday, I think I will close this PR and open a new one with Argon2id instead.

ghislainbourgeois:
Closing to move to Argon2id

Issue #192, opened 2025-02-25
Deleting an nonexistent certificate request returns 202 Accepted message


gruyaume:
#### Describe the bug

Whenever we delete a non-existent certificate request (and perhaps other items) using the HTTP API, we get a 202 Accepted return message.

#### To Reproduce

1. Create Account
2. Login
3. Delete Certificate Request

#### Expected behavior
 Here we would expect a 404 as the endpoint doesn't exist.

#### Logs
Request
```
curl --location --request DELETE 'https://127.0.0.1:3000/api/v1/certificate_requests/4' \
--header 'Authorization: ••••••' \
--data ''
```

Response:
```
{
    "result": {
        "message": "success"
    }
}
```
#### Environment

- This was tested with the version of Notary on this PR: #184 but I believe the issue was not introduced here.

kayra1:
If we delete a nonexistent certificate wouldn't the expected result still be the same :D.

But yes this is because SQL doesn't return anything if no rows matching the query were effected on delete or update operations, we will have to manually check beforehand if we want to differentiate this.

PR #203, opened 2025-03-13
chore: Improve error handling in the DB layer

saltiyazan:
# Description

This change creates error types to prevent leaking abstraction in the errors from the DB to the API layer.
It moves logging from the caller functions (API handlers) to the db functions, as by the time the error is received at the API level it will be a general error so logging it in the DB level will preserve all of the details.

# Checklist:

### DIFF
--- a/internal/db/db_certificate_authorities.go
+++ b/internal/db/db_certificate_authorities.go
@@ -9,6 +9,7 @@ import (
 	"encoding/pem"
 	"errors"
 	"fmt"
+	"log"
 	"math/big"
 	"time"
 
@@ -176,54 +177,22 @@ WITH RECURSIVE cas_with_chain AS (
 
 // ListCertificateAuthorities gets every Certificate Authority entry in the table.
 func (db *Database) ListCertificateAuthorities() ([]CertificateAuthority, error) {
-	stmt, err := sqlair.Prepare(listCertificateAuthoritiesStmt, CertificateAuthority{})
-	if err != nil {
-		return nil, err
-	}
-	var CAs []CertificateAuthority
-	err = db.conn.Query(context.Background(), stmt).GetAll(&CAs)
-	if err != nil {
-		if errors.Is(err, sqlair.ErrNoRows) {
-			return CAs, nil
-		}
-		return nil, err
-	}
-	return CAs, nil
+	return ListEntities[CertificateAuthority](db, listCertificateAuthoritiesStmt)
 }
 
 // ListDenormalizedCertificateAuthorities gets every CertificateAuthority entry in the table
 // but instead of returning ID's that reference other table rows, it embeds the row data directly into the response object.
 func (db *Database) ListDenormalizedCertificateAuthorities() ([]CertificateAuthorityDenormalized, error) {
-	stmt, err := sqlair.Prepare(listDenormalizedCertificateAuthoritiesStmt, CertificateAuthorityDenormalized{})
-	if err != nil {
-		return nil, err
-	}
-	var CAs []CertificateAuthorityDenormalized
-	err = db.conn.Query(context.Background(), stmt).GetAll(&CAs)
-	if err != nil {
-		if errors.Is(err, sqlair.ErrNoRows) {
-			return CAs, nil
-		}
-		return nil, err
-	}
-	return CAs, nil
+	return ListEntities[CertificateAuthorityDenormalized](db, listDenormalizedCertificateAuthoritiesStmt)
 }
 
 // GetCertificateAuthority gets a certificate authority row from the database.
 func (db *Database) GetCertificateAuthority(filter CertificateAuthorityFilter) (*CertificateAuthority, error) {
 	CARow, err := filter.AsCertificateAuthority()
 	if err != nil {
-		return nil, err
+		return nil, InvalidFilterError("certificate authority", err.Error())
 	}
-	stmt, err := sqlair.Prepare(getCertificateAuthorityStmt, CertificateAuthority{})
-	if err != nil {
-		return nil, err
-	}
-	err = db.conn.Query(context.Background(), stmt, CARow).Get(CARow)
-	if err != nil {
-		return nil, err
-	}
-	return CARow, nil
+	return GetOneEntity(db, getCertificateAuthorityStmt, *CARow)
 }
 
 // GetDenormalizedCertificateAuthority gets a certificate authority row from the database
@@ -232,15 +201,21 @@ func (db *Database) GetDenormalizedCertificateAuthority(filter CertificateAuthor
 	CADenormalizedRow, DenormalizedCAErr := filter.AsCertificateAuthorityDenormalized()
 	CARow, CAerr := filter.AsCertificateAuthority()
 	if CAerr != nil && DenormalizedCAErr != nil {
-		return nil, fmt.Errorf("invalid filter: only CA ID, CSR ID, or CSR PEM is supported")
+		log.Println(CAerr, DenormalizedCAErr)
+		return nil, InvalidFilterError("certificate authority", "only CA ID, CSR ID, or CSR PEM is supported")
 	}
 	stmt, err := sqlair.Prepare(getDenormalizedCertificateAuthorityStmt, CertificateAuthority{}, CertificateAuthorityDenormalized{})
 	if err != nil {
-		return nil, err
+		log.Println(err)
+		return nil, fmt.Errorf("%w: failed to get denormalized certificate authority", ErrInternal)
 	}
 	err = db.conn.Query(context.Background(), stmt, CARow, CADenormalizedRow).Get(CADenormalizedRow)
 	if err != nil {
-		return nil, err
+		if errors.Is(err, sqlair.ErrNoRows) {
+			return nil, fmt.Errorf("%w: certificate authority not found", ErrNotFound)
+		}
+		log.Println(err)
+		return nil, fmt.Errorf("%w: failed to get denormalized certificate authority", ErrInternal)
 	}
 	return CADenormalizedRow, nil
 }
@@ -275,16 +250,22 @@ func (db *Database) CreateCertificateAuthority(csrPEM string, privPEM string, ce
 	}
 	stmt, err := sqlair.Prepare(createCertificateAuthorityStmt, CertificateAuthority{})
 	if err != nil {
-		return 0, err
+		log.Println(err)
+		return 0, fmt.Errorf("%w: failed to create certificate authority", ErrInternal)
 	}
 	var outcome sqlair.Outcome
 	err = db.conn.Query(context.Background(), stmt, CARow).Get(&outcome)
 	if err != nil {
-		return 0, err
+		if isUniqueConstraintError(err) {
+			return 0, fmt.Errorf("%w: certificate authority already exists", ErrAlreadyExists)
+		}
+		log.Println(err)
+		return 0, fmt.Errorf("%w: failed to create certificate authority", ErrInternal)
 	}
 	insertedRowID, err := outcome.Result().LastInsertId()
 	if err != nil {
-		return 0, err
+		log.Println(err)
+		return 0, fmt.Errorf("%w: failed to create certificate authority", ErrInternal)
 	}
 	return insertedRowID, nil
 }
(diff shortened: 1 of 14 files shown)


internal/db/db_generic_helpers.go
kayra1, line 1: I would prefer a name like `utils` and keep the `db_` prefix for the actual db calls.

internal/db/db_certificate_authorities.go
```diff
@@ -176,37 +177,23 @@ WITH RECURSIVE cas_with_chain AS (
...
-		return nil, err
-	}
-	var CAs []CertificateAuthority
-	err = db.conn.Query(context.Background(), stmt).GetAll(&CAs)
-	if err != nil {
-		if errors.Is(err, sqlair.ErrNoRows) {
-			return CAs, nil
-		}
-		return nil, err
+		log.Println(err)
```
kayra1, line 182: We already log the whole error in the handlers. We don't need double logs for each error. Applies to all the other calls too.

internal/db/db_certificate_authorities.go
```diff
@@ -217,11 +204,15 @@ func (db *Database) GetCertificateAuthority(filter CertificateAuthorityFilter) (
...
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("%w: failed to get certificate authority", ErrInternal)
 	}
 	err = db.conn.Query(context.Background(), stmt, CARow).Get(CARow)
 	if err != nil {
-		return nil, err
+		log.Println(err)
+		if errors.Is(err, sqlair.ErrNoRows) {
+			return nil, NotFoundError("certificate authority")
```
kayra1, line 213: we should either use `fmt.Errorf()` or use custom functions everywhere for consistency. I would prefer to have the `fmt.Errorf()` since it avoids a level of abstraction.

Applies to all the other ones.

internal/db/db_certificate_authorities.go
```diff
@@ -217,11 +204,15 @@ func (db *Database) GetCertificateAuthority(filter CertificateAuthorityFilter) (
 	}
 	stmt, err := sqlair.Prepare(getCertificateAuthorityStmt, CertificateAuthority{})
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("%w: failed to get certificate authority", ErrInternal)
```
kayra1, line 207: failing to prepare the stmt is not failing to get the certificate authority, it means the SQL string is wrong. This should never fail, and if it does it's a code bug, not a user error.

applies to everywhere else.

internal/db/db_certificate_authorities.go
```diff
@@ -319,24 +329,40 @@ func (db *Database) UpdateCertificateAuthorityStatus(filter CertificateAuthority
...
 	if err != nil {
-		return err
+		log.Println(err)
+		return fmt.Errorf("%w: failed to update certificate authority", ErrInternal)
 	}
 	err = db.conn.Query(context.Background(), stmt, ca).Run()
-	return err
+	if err != nil {
+		if errors.Is(err, sqlair.ErrNoRows) {
+			return fmt.Errorf("%w: certificate authority not found", ErrNotFound)
```
kayra1, line 338: This is kind of a special case. UPDATE and DELETE sql queries will never raise a `sqlair.ErrNoRows` even if the updated/deleted rows are 0. The fix for this should be in a different PR related to #192 so I'm ok with not changing it here.

internal/db/db_certificates.go
```diff
@@ -98,15 +92,15 @@ func (db *Database) AddCertificateChainToCertificateRequest(csrFilter CSRFilter,
 	}
 	err = ValidateCertificate(certPEM)
 	if err != nil {
-		return 0, errors.New("cert validation failed: " + err.Error())
+		return 0, ErrInvalidCertificate
```
kayra1, line 95: the error returned from `ValidateCertificate` is also relevant, as it specifies the actual error (wrong chain, asn1 encoding error etc.). It should be added to the end.

internal/db/db_csrs.go
```diff
@@ -159,72 +160,42 @@ FROM certificate_chain
...
+		return nil, fmt.Errorf("%w: failed to list certificate requests", ErrInternal)
 	}
 	return csrs, nil
 }
 
+// ListCertificateRequests gets every CertificateRequest entry in the table.
+func (db *Database) ListCertificateRequests() ([]CertificateRequest, error) {
+	return listCertificateRequests(db, listCertificateRequestsStmt)
+}
+
```
kayra1, line 185: I don't like that this breaks the pattern we have with every other db call in the other files. It won't hurt to repeat this function twice.

internal/db/db_users.go
```diff
@@ -79,25 +77,39 @@ func (db *Database) GetUser(filter UserFilter) (*User, error) {
...
+		return 0, fmt.Errorf("%w: failed to create user", ErrInternal)
 	}
 	row := User{
 		Username:       username,
 		HashedPassword: pw,
 		Permissions:    permission,
 	}
+	err = ValidateUser(row)
+	if err != nil {
+		return 0, err
```
kayra1, line 98: we should use `ErrInvalidInput` here like all the other spots.

internal/db/errors.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+)
+
+var (
+	ErrNotFound                  = errors.New("resource not found")
+	ErrInternal                  = errors.New("internal error")
```
kayra1, line 11: lets use ErrInternal and include the error statement returned for all of the `sqlair.Prepare` errors.

internal/db/errors.go
```diff
@@ -0,0 +1,36 @@
...
+
+// NotFoundError customizes the error message for not found errors
+func NotFoundError(entityName string) error {
+	return fmt.Errorf("%w: %s", ErrNotFound, entityName)
+}
+
+// InvalidFilterError customizes the error message for invalid filter errors
+func InvalidFilterError(filterType, message string) error {
+	return fmt.Errorf("%w: %s - %s", ErrInvalidFilter, filterType, message)
+}
```
kayra1, line 27: Like I said above, let's just directly use `fmt.Errorf`. we already use it almost everywhere.

internal/db/validation.go
```diff
@@ -122,6 +122,19 @@ func ValidatePrivateKey(pk string) error {
...
+	}
+	if user.HashedPassword == "" {
+		return fmt.Errorf("%w: invalid username or password", ErrInvalidInput)
+	}
+	if user.Permissions != 0 && user.Permissions != 1 {
+		return fmt.Errorf("%w: invalid permissions", ErrInvalidInput)
+	}
+	return nil
+}
+
```
kayra1, line 137: We already validate exactly this [here](https://github.com/canonical/notary/blob/ca702a67d3aa9b9acae652e3a0c2f906b86521b4/internal/server/handlers_accounts.go#L22), and I've avoided validating this twice, but it seems to keep cropping up in PR's so I'm ok with just finally merging it.

internal/server/handlers_accounts.go
```diff
@@ -117,8 +114,7 @@ func GetAccount(env *HandlerConfig) http.HandlerFunc {
 			account, err = env.DB.GetUser(db.ByUserID(idNum))
 		}
 		if err != nil {
-			log.Println(err)
-			if errors.Is(err, sqlair.ErrNoRows) {
+			if errors.Is(err, db.ErrNotFound) || errors.Is(err, db.ErrInvalidFilter) {
```
kayra1, line 117: If the filter is wrong that's an internal error, but it's clearly not possible to get that error here through user error, it would be a code bug. I'm not sure if we should keep it like this, remove checking for invalid filter, or split it up and introduce a code path that's not possible to get.

internal/server/handlers_certificate_requests.go
```diff
@@ -142,22 +138,19 @@ func GetCertificateRequest(env *HandlerConfig) http.HandlerFunc {
...
 			return
 		}
 		_, err = env.DB.GetCertificateAuthority(db.ByCertificateAuthorityCSRID(csr.CSR_ID))
-		if rowFound(err) {
-			log.Println(err)
-			writeError(w, http.StatusNotFound, "Not Found")
-			return
-		}
-		if realError(err) {
-			log.Println(err)
```
kayra1, line 160: The reason this was like this was because of Guillaume's preference, but I'm happy with either version.

kayra1 (changes requested):
Just some stuff across all the calls but overall great work, thanks.

internal/db/db_generic_helpers.go
saltiyazan: ✅

internal/db/db_certificate_authorities.go
saltiyazan: I am moving them from the handlers here, the handlers will only see very general errors now, so if we don't want to lose the information we have to log earlier.

internal/db/db_certificate_authorities.go
saltiyazan: True, it is checked however at the beginning of the function when we try to get the current record based on the filter, if it isn't there we return.

internal/db/db_certificate_authorities.go
saltiyazan: Done

internal/db/db_csrs.go
saltiyazan: Reverted.

internal/db/db_users.go
saltiyazan: ✅

internal/db/errors.go
saltiyazan: ✅

internal/db/validation.go
saltiyazan: There are constraints on the db level too, while validating it on an API level should be a good enough guard, I think we can validate on a db level too.

internal/server/handlers_accounts.go
saltiyazan: I removed it since we can't really get there.

internal/server/handlers_accounts_test.go
```diff
@@ -285,14 +285,13 @@ func TestAccountsEndToEnd(t *testing.T) {
...
 		}
-		if statusCode != http.StatusNotFound {
-			t.Fatalf("expected status %d, got %d", http.StatusNotFound, statusCode)
+		if statusCode != http.StatusUnauthorized {
+			t.Fatalf("expected status %d, got %d", http.StatusUnauthorized, statusCode)
 		}
-		if response.Error != "Not Found" {
-			t.Fatalf("expected error %q, got %q", "Not Found", response.Error)
+		if response.Error != "Unauthorized" {
+			t.Fatalf("expected error %q, got %q", "Unauthorized", response.Error)
```
kayra1, line 292: Why is this unauthorised now? we have provided the admin token here, it should return a user not found since it's trying to change a password for an account that doesn't exist.

kayra1 (dismissed):

internal/server/handlers_accounts.go
```diff
@@ -270,9 +262,8 @@ func ChangeAccountPassword(env *HandlerConfig) http.HandlerFunc {
 		}
 		err = env.DB.UpdateUserPassword(db.ByUserID(idNum), changeAccountParams.Password)
 		if err != nil {
-			log.Println(err)
-			if errors.Is(err, sqlair.ErrNoRows) {
-				writeError(w, http.StatusNotFound, "Not Found")
+			if errors.Is(err, db.ErrNotFound) {
+				writeError(w, http.StatusUnauthorized, "Unauthorized")
```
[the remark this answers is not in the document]
kayra1: We are explicitly allowing admin users to be able to change other users' passwords so we don't want this behaviour. Even if we did, this is not the place to change permissions. They are handled in the middleware where if the user doesn't have permission to use an endpoint they will always receive a 403 Forbidden. Unauthorised also means the user isn't logged in, which is a different thing than what's happening here. The verb you'd want would have been 403 forbidden.

internal/db/db_certificate_authorities.go
kayra1: What I mean with this comment is we should reflect that in the text, and instead of saying failed to get certificate authority we should say something like SQL compilation failed or something along those lines.

internal/db/db_certificate_authorities.go
kayra1: in that case we don't need to have the ErrNoRows check at all

saltiyazan: Done in d41c1da3474792c3c20423ffa31df34fa61a74b7
This means that #192 is addressed as a consequence in this PR.

internal/server/handlers_accounts.go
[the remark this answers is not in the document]
saltiyazan: Okay I reverted the change in d7cffe6a7ef89f13f279b16f23fdb8b5300cde7a

internal/db/db_certificate_authorities.go
```diff
@@ -217,11 +204,15 @@ func (db *Database) GetCertificateAuthority(filter CertificateAuthorityFilter) (
 	}
 	stmt, err := sqlair.Prepare(getCertificateAuthorityStmt, CertificateAuthority{})
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("%w: failed to get certificate authority due to sql compilation error", ErrInternal)
```
kayra1, line 207: can we also add the err to the message at the end:
`"%w: failed to get certificate authority due to sql compilation error: %w", ErrInternal, err`

applies to all the other ones

kayra1 (changes requested):
Great job, one last nit

kayra1 (approved):

internal/db/db_certificate_authorities.go
saltiyazan: I added log statements :)

PR #204, opened 2025-03-17
feat: certificate revocation through CRL

kayra1:
# Description

This change introduces CRL's into the CA's in notary. CA certificates will be generated with a CRL embedded into the DB row, and the CA certificate will be created with a CRLDistributionPoint extension that points to `api/v1/certificate_authorities/{id}/crl`. This endpoint will return the CRL of the certificate. This endpoint is open to unauthenticated users.

Revoked CA or end certificates' serial numbers will be placed in the CRL before being removed from notary. revoking a certificate that wasn't signed by notary will no longer be allowed, since notary will not have the authority to do that.

When revoking an intermediate certificate, the final CRL will remain in the system for past its expiry date, since a certificate is required to sign the CRL. However, revoking the intermediate's issued certificates is left to the user themselves. RFC 5280 does not specify that the certificates issued by intermediate certificates should also be revoked, so we need to support cases where users may want to revoke an intermediate but continue using its issued certificates until they have another intermediate available.

# Checklist:


internal/db/validation.go
```diff
@@ -153,3 +156,15 @@ func sanitizeCertificateBundle(cert string) ([]string, error) {
...
+func HashPassword(password string) (string, error) {
+	if strings.TrimSpace(password) == "" {
+		return "", fmt.Errorf("%w: password cannot be empty", ErrInvalidInput)
+	}
+	hashedPassword, err := bcrypt.GenerateFromPassword([]byte(password), bcrypt.DefaultCost)
+	if err != nil {
+		return "", err
+	}
+	return string(hashedPassword), nil
+}
```
ghislainbourgeois, line 170: This was removed previously, probably a conflict resolution gone wrong?

internal/db/db_csrs_test.go
```diff
@@ -211,14 +211,14 @@ func TestRevokeCertificateRequestFails(t *testing.T) {
...
-		t.Fatalf("Expected failure revoking nonexistent CSR")
-	}
+	// err = database.RevokeCertificate(db.ByCSRPEM("this is definitely not a csr"))
+	// if err == nil {
+	// 	t.Fatalf("Expected failure revoking nonexistent CSR")
+	// }
+	// err = database.RevokeCertificate(db.ByCSRID(-1)) TODO
+	// if err == nil {
+	// 	t.Fatalf("Expected failure revoking nonexistent CSR")
+	// }
```
ghislainbourgeois, line 221: This should be uncommented, or deleted.

internal/db/db_csrs_test.go
```diff
@@ -263,14 +263,14 @@ func TestCASNotShowingUpInCSRsTable(t *testing.T) {
...
-		t.Fatalf("Couldn't create certificate authority: %s", err)
-	}
+	// _, err = database.CreateCertificateAuthority(RootCACSR, RootCAPrivateKey, RootCACertificate+"\n"+RootCACertificate)
+	// if err != nil {
+	// 	t.Fatalf("Couldn't create certificate authority: %s", err)
+	// }
+	// _, err = database.CreateCertificateAuthority(IntermediateCACSR, IntermediateCAPrivateKey, "")
+	// if err != nil {
+	// 	t.Fatalf("Couldn't create certificate authority: %s", err)
+	// }
```
ghislainbourgeois, line 273: Same here

internal/server/handlers_certificate_authorities.go
```diff
@@ -190,28 +195,38 @@ func createCertificateAuthority(fields CreateCertificateAuthorityParams) (string
...
 	if err != nil {
-		return "", "", ""
+		return "", "", "", "", fmt.Errorf("error creating certificate authority: %w", err)
 	}
-
-	return csrPEM.String(), privPEM.String(), certPEM.String()
+	crlBytes, err := x509.CreateRevocationList(rand.Reader, &x509.RevocationList{
+		Number:     big.NewInt(time.Now().UnixNano()),
+		ThisUpdate: time.Now(),
+		NextUpdate: time.Now().AddDate(nextUpdateYears, 0, 0),
```
ghislainbourgeois, line 219: Is there a way that we ensure that there is a new CRL generated before `NextUpdate`? To respect the RFC, we need to issue a new one (even if there are no new certificates added to the list) no later than that time.

ghislainbourgeois (changes requested):
I did an incomplete first pass, overall it looks good, but I had some small comments.

internal/server/handlers_certificate_authorities.go
kayra1: I will handle this and all expiry/renewal events in TLSENG-594

internal/db/db_csrs_test.go
kayra1: uncommented

internal/db/db_csrs_test.go
kayra1: uncommented

internal/config/config.go
```diff
@@ -50,6 +52,9 @@ func Validate(filePath string) (Config, error) {
 	if err != nil {
 		return Config{}, err
 	}
+	if c.ExternalHostname == "" {
+		return Config{}, errors.New("`external_hostname` is empty")
+	}
```
gruyaume, line 57: Why force an external hostname? For small or dev environments, we won't have a dns.

internal/server/utils.go
```diff
@@ -0,0 +1,26 @@
...
+		_                asn1.RawValue
+		SubjectPublicKey asn1.BitString
+	}
+	_, err := asn1.Unmarshal(pubKeyBytes, &spki)
+	if err != nil {
+		return nil, err
+	}
+	hash := sha1.Sum(spki.SubjectPublicKey.Bytes)
+	return hash[:], nil
+}
```
gruyaume, line 26: We should handle the case where `priv` is nil.

internal/server/utils.go
```diff
@@ -0,0 +1,26 @@
...
+	"crypto/x509"
+	"encoding/asn1"
+)
+
+// generateSKI generates the Subject Key Identifier (SKI) for the given private key.
+// The SKI is the SHA-1 hash of the public key.
+// The SKI is used to identify the public key in the certificate and is used in the Authority Key Identifier extension.
+// The SKI is necessary for CRL signing.
+func generateSKI(priv *rsa.PrivateKey) ([]byte, error) {
+	pubKeyBytes, _ := x509.MarshalPKIXPublicKey(&priv.PublicKey)
```
gruyaume, line 15: Why ignore the error coming out of `MarshalPKIXPublicKey`? We should likely capture it and return it

internal/server/utils.go
```diff
@@ -0,0 +1,26 @@
...
+	var spki struct {
+		_                asn1.RawValue
+		SubjectPublicKey asn1.BitString
+	}
+	_, err := asn1.Unmarshal(pubKeyBytes, &spki)
+	if err != nil {
+		return nil, err
+	}
```
gruyaume, line 23: Generally speaking, it feels like we could do a better job at error handling. Here for instance we're not wrapping the error, making the end log not capture the fact that what we were trying to unmarshall is a public key. 

Suggestion:
```go
if err != nil {
    return nil, fmt.Errorf("failed to unmarshal public key: %w", err)
}
```

I think this comment applies to a bunch of places in this project. I don't think we should go ahead and refactor it everywhere right now - but in general we should be careful to produce meaningful logs.

internal/server/utils.go
gruyaume, line 1: We should probably add unit test(s) for this function

internal/db/utils.go
```diff
@@ -42,3 +48,90 @@ func GetOneEntity[T any](db *Database, query string, params T) (*T, error) {
...
 	return &result, nil
 }
+
+// ParseCertificateChain receives a PEM string chain and returns an x.509.Certificate list.
+func ParseCertificateChain(pemChain string) ([]*x509.Certificate, error) {
+	var certs []*x509.Certificate
+	certChainStr, err := SplitCertificateBundle(pemChain)
+	if err != nil {
+		return nil, err
+	}
```
gruyaume, line 58: Please look at the comment I gave in `internal/server/utils.go` regarding error wrapping. This comment applies at a couple of places in this file.

gruyaume (changes requested):
I want to run the change on my machine to try it for myself. For now I mostly focused on non-functional aspects of the changes. My main concerns are:
- We could improve error handling
- We shouldn't make `external_hostname` a mandatory configuration option

internal/config/config.go
kayra1: should I default the value to localhost in that case or just allow empty values? This field is only used for the CRLDistributionPoints.

internal/server/utils.go
kayra1: This function is an internal private function that isn't expected to be used anywhere except for the only other also internal function in `handlers_certificate_authorities.go`. That's why I discard errors here and don't handle inputs that are not possible. I can convert this into a public function and write tests and such, but for a private function it would be statically impossible to provide a nil value to this function in regular operation. I avoid handling errors that are statically impossible usually, and `golangci-lint` also does since it doesn't raise an errcheck lint error for this.

internal/server/utils.go
kayra1: same reason as [this](https://github.com/canonical/notary/pull/204#discussion_r2016341916) comment.

internal/server/utils.go
kayra1: I agree and I've slowly started tackling this issue, yazan's PR [here](https://github.com/canonical/notary/pull/203) being the first attempt in the database layer. I've opened TSLENG-689 for this issue.

internal/db/utils.go
kayra1: let's continue the discussion there.

internal/server/utils.go
kayra1: Go doesn't allow importing internal functions into test packages. I can make this a public function.

internal/config/config.go
kayra1: This is what I've done for now but I can change the default value to anything.

ghislainbourgeois (approved):
LGTM

docs/reference/api/certificate_authorities.md
```diff
@@ -83,7 +83,41 @@ None
 }
 ```
 
-## Update a Certificate Authority
+## Get the CRL of a Certificate Authority
+
+This path returns a single certificate authority.
```
gruyaume, line 88: Not quite

internal/server/handlers_certificate_authorities.go
```diff
@@ -478,3 +495,70 @@ func SignCertificateAuthority(env *HandlerConfig) http.HandlerFunc {
...
+			writeError(w, http.StatusInternalServerError, "internal error")
+			return
+		}
+	}
+}
+
+// RevokeCertificateAuthorityCertificate handler receives an id as a path parameter,
+// and revokes the corresponding certificate by placing the certificate serial number to the CRL
+// It returns a 200 OK on success
+func RevokeCertificateAuthorityCertificate(env *HandlerConfig) http.HandlerFunc {
```
gruyaume, line 533: Notary won't be able to revoke a CA that was signed outside of it - we should probably return a specific error when someone tries to do this.

docs/reference/api/certificate_requests.md
```diff
@@ -121,13 +121,38 @@ This path creates a certificate for a certificate request.
 }
 ```
 
-## Reject a Certificate for a Certificate Request
+## Reject a Certificate Request
+
+This path rejects a certificate request. This has no meaning in the TLS ecosystem, 
+and is used for users to easily differentiate outstanding certificates.
```
gruyaume, line 127: I'm not sure about the "This has no meaning in the TLS ecosystem". It doesn't really help us understand what this API call actually does

docs/reference/api/certificate_requests.md
```diff
@@ -121,13 +121,38 @@ This path creates a certificate for a certificate request.
...
+
+```json
+{
+    "result": {
+        "message": "success"
+    }
+}
+```
+
+## Revoke a Certificate from a Certificate Request
```
gruyaume, line 147:
```suggestion
## Revoke a Certificate
```

docs/reference/api/certificate_requests.md
```diff
@@ -121,13 +121,38 @@ This path creates a certificate for a certificate request.
...
+
+This path revokes an existing certificate from a certificate request. This path only works if the certificate request was signed in notary.
+Notary will place the certificate's serial number in the CRL of the issuing CA.
```
gruyaume, line 150:
```suggestion
This path revokes an existing certificate. This path only works if the certificate request was signed in notary.
Notary will place the certificate's serial number in the CRL of the issuing CA.
```

internal/server/handlers_certificate_requests.go
```diff
@@ -346,7 +346,7 @@ func DeleteCertificate(env *HandlerConfig) http.HandlerFunc {
 }
 
 // RevokeCertificate handler receives an id as a path parameter,
-// and attempts to revoke the corresponding certificate request
+// and attempts to revoke the corresponding certificate request by adding the certificate to the CRL
 // It returns a 200 OK on success
 func RevokeCertificate(env *HandlerConfig) http.HandlerFunc {
```
gruyaume, line 351: Notary won't be able to revoke a certificate that was signed outside of it - we should probably return a specific error when someone tries to do this.

internal/server/utils.go
gruyaume: > This function is an internal private function that isn't expected to be used anywhere except for the only other also internal function in handlers_certificate_authorities.go

That's the case currently but the function is accessible to anything under `internal/server`. Tomorrow, someone else could use it elsewhere in that same part of the project with a `non-nil` private key. 

Also, why place it under the `utils` folder if it's only expected to be used in 1 place? 

> and golangci-lint also does since it doesn't raise an errcheck lint error for this.

Try and add call `generateSKI` with a `nil` value. You will notice that golangci-lint doesn't error out.

internal/server/utils.go
gruyaume: I still don't get the added value of not checking for the error. The callee doesn't have information about the caller, why assume anything about it?

gruyaume (changes requested):

internal/server/utils.go
kayra1: > That's the case currently but the function is accessible to anything under internal/server. Tomorrow, someone else could use it elsewhere in that same part of the project with a non-nil private key.

There is no way to prevent this function or any function from being used in the same namespace. Another developer can easily break any function if they use it wrong, 

> Also, why place it under the utils folder if it's only expected to be used in 1 place?

It is a utility/helper function, one that makes the calling function more readable. I hadn't considered the number of uses when deciding where to put functions. The way I approach that is "would the developer expect to find a function of this name in this file?" I would personally only expect to find certificate authority handler functions in the `handlers_certificate_authorities.go` file.

> Try and add call generateSKI with a nil value. You will notice that golangci-lint doesn't error out.

exactly. golangci-lint does not consider it bad to not check for errors here because it doesn't error out.


---


The spec PL007 specifies how to handle cases like this, where the caller uses invalid arguments or what to do with unreachable code. It recommends to add a simple panic statement. I think that's a good balance between ignoring errors and going through a bunch of effort and creating a bunch of noise to categorize the exact error type and creating messages. I've used it here, let me know what you think.

internal/server/utils.go
kayra1: not checking for the error is the null value, where we discard errors that wouldn't effect the program (in this case because this is unreachable code). I don't see the value in having detailed and informative error messages and handling in code that people can't see, even if that may be possible in the future. I would consider it the responsibility of the developer to not misuse internal functions.

justifications aside, I think there's a better common ground way to do this, I've responded to it above.

internal/server/handlers_certificate_authorities.go
```diff
@@ -20,12 +20,19 @@ import (
...
 
+const nextUpdateYears = 1
+
 type CertificateAuthority struct {
 	ID             int64       `json:"id"`
 	Status         db.CAStatus `json:"status"`
 	PrivateKeyPEM  string      `json:"private_key,omitempty"`
 	CertificatePEM string      `json:"certificate"`
 	CSRPEM         string      `json:"csr"`
+	CRL            string      `json:"crl"`
```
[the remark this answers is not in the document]
kayra1: When getting a certificate authority I thought users would expect to be able to see everything we have available about the object, and the specific `/<id>/crl` endpoint is for the CRLDistributionPoint extension and to simplify automation with other tooling.

docs/reference/api/certificate_authorities.md
kayra1: :(

docs/reference/api/certificate_requests.md
kayra1: I didn't know how else to phrase it. I tried again.

internal/server/handlers_certificate_authorities.go
kayra1: Added: "tcertificates need to be signed by a notary managed certificate authority in order to be revoked"

internal/server/handlers_certificate_requests.go
kayra1: added the same message as [my comment above](https://github.com/canonical/notary/pull/204#discussion_r2026397449).

gruyaume (approved):
Approved, assuming CI issues related to spellcheck gets fixed

PR #211, opened 2025-03-18
feat: Use argon2id for password hashing as recommended by OWASP

ghislainbourgeois:
# Description

This PR replaces the use of `bcrypt` by `argon2id` as recommended by [OWASP](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#password-hashing-algorithms). It uses the default parameters recommended by the same organization and creates a struct that can be used in the future to configure the parameters.

Because the Go library for `argon2id` is less complete than `bcrypt`, I also had to implement the salt generation, and the PHC formatting.

I took the opportunity to refactor a bit the login operation, so that when a user is not found, we still hash the password, so that an attacker cannot extract information about existing users by analyzing the timing.

# Checklist:

### DIFF
--- a/internal/hashing/argon2id.go
+++ b/internal/hashing/argon2id.go
@@ -46,7 +46,7 @@ func HashPassword(password string) (string, error) {
 
 	encoded_salt := base64.RawStdEncoding.EncodeToString(salt)
 	encoded_hash := base64.RawStdEncoding.EncodeToString(hash)
-	hashedPassword := fmt.Sprintf("$argon2id$v=%d$m=%d,t=%d,p=%d$%s$%s",
+	hashedPassword := fmt.Sprintf(phcFormat,
 		argon2.Version,
 		DefaultArgon2IDParameters.Memory,
 		DefaultArgon2IDParameters.Time,
--- a/internal/hashing/argon2id.go
+++ b/internal/hashing/argon2id.go
@@ -6,13 +6,15 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
-	"strconv"
+	"math"
 	"strings"
 
 	"golang.org/x/crypto/argon2"
 )
 
-const phcFormat = "$argon2id$v=%d$m=%d,t=%d,p=%d$%s$%s"
+const phcEncFormat = "$argon2id$v=19$m=%d,t=%d,p=%d$%s$%s"
+
+var phcDecFormat = strings.ReplaceAll(phcEncFormat, "$", " ")
 
 type Argon2IDParameters struct {
 	SaltLength uint32 // bytes
@@ -20,6 +22,22 @@ type Argon2IDParameters struct {
 	Memory     uint32 // kibibytes
 	Threads    uint8  // number of threads
 	KeyLength  uint32 // bytes
+}
+
+type phcEncoding struct {
+	params Argon2IDParameters
+	salt   []byte
+	key    []byte
+}
+
+func (e *phcEncoding) String() string {
+	return fmt.Sprintf(phcEncFormat,
+		e.params.Memory,
+		e.params.Time,
+		e.params.Threads,
+		base64.RawStdEncoding.EncodeToString(e.salt),
+		base64.RawStdEncoding.EncodeToString(e.key),
+	)
 }
 
 // Default parameters recommended by OWASP
@@ -35,26 +53,18 @@ var ErrInvalidPassword = errors.New("invalid password")
 
 // Takes the password string, makes sure it's not empty, and hashes it using argon2id
 func HashPassword(password string) (string, error) {
+	var err error
 	if strings.TrimSpace(password) == "" {
 		return "", fmt.Errorf("%w: password cannot be empty", ErrInvalidPassword)
 	}
-	salt, err := generateSalt(DefaultArgon2IDParameters.SaltLength)
+	encoding := phcEncoding{params: DefaultArgon2IDParameters}
+	encoding.salt, err = generateSalt(DefaultArgon2IDParameters.SaltLength)
 	if err != nil {
 		return "", err
 	}
-	hash := hashPasswordWithSaltAndParams(password, salt, DefaultArgon2IDParameters)
+	encoding.key = hashPasswordWithSaltAndParams(password, encoding.salt, encoding.params)
 
-	encoded_salt := base64.RawStdEncoding.EncodeToString(salt)
-	encoded_hash := base64.RawStdEncoding.EncodeToString(hash)
-	hashedPassword := fmt.Sprintf(phcFormat,
-		argon2.Version,
-		DefaultArgon2IDParameters.Memory,
-		DefaultArgon2IDParameters.Time,
-		DefaultArgon2IDParameters.Threads,
-		encoded_salt,
-		encoded_hash,
-	)
-	return string(hashedPassword), nil
+	return encoding.String(), nil
 }
 
 // CompareHashAndPassword takes a hashed password string and a password,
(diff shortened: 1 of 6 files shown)


internal/hashing/argon2id.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+// CompareHashAndPassword takes a hashed password string and a password,
+// hashes the password with the same parameters as the hashed password and
+// compares the resulting value, returning an error if they do not match.
+//
+// When this function is passed an invalid hashed password string, it will
+// hash the password with the default parameters to prevent an attacker from
+// getting information from the timing of a login failure.
+func CompareHashAndPassword(hashedPassword string, password string) error {
```
gruyaume, line 66: For reference, here are some upstream discussions on adding these helpers for argon2id:
- https://github.com/golang/go/issues/60740
- https://github.com/golang/go/issues/16971

Someone tried to add this feature to the upstream through the PR referenced below, but never got a review from the maintainers:
- https://go-review.googlesource.com/c/crypto/+/431595

I don't know if you looked at the implementation there, maybe we can use some of those ideas.

internal/hashing/argon2id.go
```diff
@@ -0,0 +1,162 @@
...
+)
+
+const ARGON2_VERSION = 19
+
+type Argon2IDParameters struct {
+	SaltLength uint32
+	Time       uint32
+	Memory     uint32
+	Threads    uint8
+	KeyLength  uint32
```
gruyaume, line 21: Should we specify the units next to the parameters? Ex:

```go

type Argon2IDParameters struct {
	SaltLength uint32 // bytes
}
```

internal/hashing/argon2id.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+	salt, err := generateSalt(DefaultArgon2IDParameters.SaltLength)
+	if err != nil {
+		return "", err
+	}
+	hash := hashPasswordWithSaltAndParams(password, salt, DefaultArgon2IDParameters)
+
+	encoded_salt := base64.RawStdEncoding.EncodeToString(salt)
+	encoded_hash := base64.RawStdEncoding.EncodeToString(hash)
+	hashedPassword := fmt.Sprintf("$argon2id$v=%d$m=%d,t=%d,p=%d$%s$%s",
```
gruyaume, line 48: I think it would be simpler to have an encode and decode function for the hashed password.

```
func encode(params Argon2IDParameters) (string, error)
func decode(encoded string) (*Argon2IDParameters, error) 
```

They seem to do something similar [here](https://go-review.googlesource.com/c/crypto/+/431595/1/argon2/encoding.go#48)

gruyaume, line 48: We could use a const for  `argon2id` as we use it here to encode, and we use it in the docoder code as well.

internal/hashing/argon2id.go
```diff
@@ -0,0 +1,162 @@
...
+
+// Parses a PHC string for Argon2id, and returns Argon2id parameters, the salt and the hash
+func parseArgon2IDPHC(phc string) (Argon2IDParameters, []byte, []byte, error) {
+	fields := strings.Split(phc, "$")
+	if len(fields) != 6 {
+		return DefaultArgon2IDParameters, nil, nil, fmt.Errorf("cannot parse hashed password strings for argon2id")
+	}
+	if fields[1] != "argon2id" || fields[2] != fmt.Sprintf("v=%d", ARGON2_VERSION) {
+		return DefaultArgon2IDParameters, nil, nil, fmt.Errorf("cannot parse hashed password strings for argon2id")
+	}
```
gruyaume, line 109: I would encourage to separate these into 2 conditions and have more specific error messages (ex. `mode is not argondID: <whatever mode was found>`, `version not 19: <whatever version was found>`.

internal/hashing/argon2id.go
```diff
@@ -0,0 +1,162 @@
...
+	"strconv"
+	"strings"
+
+	"golang.org/x/crypto/argon2"
+)
+
+const ARGON2_VERSION = 19
+
+type Argon2IDParameters struct {
+	SaltLength uint32
```
gruyaume, line 17: Why don't we include the version in the params?

gruyaume (changes requested):

internal/hashing/argon2id.go
ghislainbourgeois: I did not find those discussions when I did my researched, but found the third party library that is mentioned in the first link. Overall it looks similar, but I think the encoding/decoding idea is a good one and adding support for the `pepper` is also interesting.

internal/hashing/argon2id.go
ghislainbourgeois: Yes, I think that would be useful. I will add that.

internal/hashing/argon2id.go
ghislainbourgeois: It was not in there because it is basically the only supported version by Go and made no sense as a parameter, but with the encoding changes, it might make sense to bring it in.

internal/hashing/argon2id.go
ghislainbourgeois: I went for a single error message here because this is a package scoped function that is only used in a single place and we do not do anything with the error. I used simple `fmt.Errorf` as those are unrecoverable errors, and followed PL007 here.

I am not against making the different errors more explicit, but I doubt there is value here, as they should not be logged (this code will get called on every login, including with invalid users, so an attacker could fill out the logs with a DoS attack), and the outcome is the same in all cases; we cannot decode the hash, so the password will not match.

kayra1 (changes requested):
looks good, thanks. just 1 thing.

internal/server/handlers_login.go
```diff
@@ -68,14 +68,16 @@ func Login(env *HandlerConfig) http.HandlerFunc {
 		}
 		userAccount, err := env.DB.GetUser(db.ByUsername(loginParams.Username))
 		if err != nil {
-			if errors.Is(err, db.ErrNotFound) || errors.Is(err, db.ErrInvalidFilter) {
-				writeError(w, http.StatusUnauthorized, "The username or password is incorrect. Try again.")
+			if !errors.Is(err, db.ErrNotFound) && !errors.Is(err, db.ErrInvalidFilter) {
```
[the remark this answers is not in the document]
ghislainbourgeois: I just refactored this code, it was added recently, what should I do with it instead? My guess would be a 500, as this in an internal error, right?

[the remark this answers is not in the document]
kayra1: Oh I see, let's keep it as is then if there are other places that this exists.

internal/hashing/argon2id.go
ghislainbourgeois: I took some inspiration from those, like using `crypto/subtle` and using `Strict()` for base64 decoding.

internal/hashing/argon2id.go
ghislainbourgeois: I added the units.

internal/hashing/argon2id.go
ghislainbourgeois: I moved this to a constant.

internal/hashing/argon2id.go
ghislainbourgeois: I changed to specific errors.

internal/server/handlers_login.go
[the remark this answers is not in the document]
ghislainbourgeois: @kayra1 do you want me to do something with this, or should I resolve the discussion with no action?

internal/hashing/argon2id.go
ghislainbourgeois: I also removed the version constant from this file and instead used the one from the standard library.

internal/hashing/argon2id.go
ghislainbourgeois: I now use a similar approach. I kept the public struct `Argon2IDParameters` separate, as this will eventually be used to configure the hashing parameters. Internally, I added the `phcEncoding` struct.

kayra1 (approved):

gruyaume (approved):

PR #227, opened 2025-04-09
chore: refactor test code in notary (part 1)

kayra1:
# Description

This PR is the first part in a series of PR's that aim to unify the notary backend tests into a cohesive code style that is recommended by the google style guide, best practices and the spec PL007. The actual changes are as follows:

## config
* started using t.Run for all tests
* avoid using TestMain in favor of helper functions
* used cmp for validating complex struct returns
* standardized test error messages with `got` and `want`
* ordered functions as tests, helpers, consts to make the file easier to read top to bottom.
* made debug logging to stdout default, so logging is no longer required to be defined in the config.

## database & server
* moved helpers into a common testutils package.
* common helper functions that set up the database

Since this PR just reorganizes the tests, it has a high LOC count and a net -300 lines of code. The next PR will focus on improving the actual tests themselves, even though the changes have reduced the total test time by 16%.

# Checklist:


internal/server/handlers_login_test.go
```diff
@@ -85,30 +41,24 @@ func TestLoginEndToEnd(t *testing.T) {
 		if loginResponse.Result.Token == "" {
 			t.Fatalf("expected token, got empty string")
 		}
-		token, err := jwt.Parse(loginResponse.Result.Token, func(token *jwt.Token) (interface{}, error) {
-			if _, ok := token.Method.(*jwt.SigningMethodHMAC); !ok {
-				return nil, fmt.Errorf("Unexpected signing method: %v", token.Header["alg"])
-			}
-			return config.JWTSecret, nil
-		})
+		// TODO: consider better jwt options: RFC 7519, 7517, 7515
```
saltiyazan, line 44: Is this TODO in the right place here? Can we create an issue instead?

internal/testutils/dbtestutils.go
```diff
@@ -0,0 +1,42 @@
...
+	t.Cleanup(func() {
+		database.Close()
+	})
+	return database
+}
+
+func MustPrepareMockDB(t *testing.T) {
+	t.Helper()
+	// create a db with mock data
+}
```
saltiyazan, line 30: Should we add this whenever it becomes needed?

internal/testutils/testdata.go
saltiyazan, line 1: Why don't util files have underscores in their names like the rest of the files?

internal/testutils/servertestutils.go
```diff
@@ -0,0 +1,785 @@
...
+	if err != nil {
+		t.Fatalf("couldn't login admin account: %s", err)
+	}
+	if statusCode != http.StatusOK {
+		t.Fatalf("expected status %d, got %d", http.StatusOK, statusCode)
+	}
+	return loginResponse.Result.Token
+}
+
+func MustPrepareNonAdminAccount(t *testing.T, ts *httptest.Server, adminToken string) string {
```
saltiyazan, line 74:
```suggestion
func MustPrepareNonAdminAccount(t *testing.T, ts *httptest.Server, token string) string {
```

internal/testutils/servertestutils.go
```diff
@@ -0,0 +1,785 @@
...
+		t.Fatalf("Couldn't get server: %s", err)
+	}
+	testServer := httptest.NewTLSServer(notaryServer.Handler)
+	t.Cleanup(func() {
+		testServer.Close()
+	})
+	return testServer
+}
+
+func MustPrepareAdminAccount(t *testing.T, ts *httptest.Server) string {
```
saltiyazan, line 46: This will create the so called `testadmin` user when it is called even if a user already exists in the PR, in that case the user created won't be admin. Or am I mistaken?

saltiyazan: It only differs from the other function by the username and the password and whether it creates an admin user or a normal user depends on how it is called.

internal/metrics/metrics_test.go
saltiyazan, line 1: Can't this file use the shared test data file for its mock certs?

internal/db/db_init.go
```diff
@@ -15,17 +16,17 @@ import (
 
 // Database is the object used to communicate with the established repository.
 type Database struct {
-	conn          *sqlair.DB
```
saltiyazan, line 18: Do you think we should make this public just for the sake of tests? I'm not saying we shouldn't, but I want to raise the point.

saltiyazan (changes requested):

internal/server/handlers_login_test.go
kayra1: I will address this in the next PR, i removed it for now but didn't create a new issue, just wrote it in a text file locally.

internal/testutils/testdata.go
kayra1: i don't know :D fixed

internal/metrics/metrics_test.go
kayra1: good catch, removed those

internal/db/db_init.go
kayra1: I think it's worth it. This is an internal package anyway, and we may have legitimate reasons to expose it too. I'm also more open to just adding the tests into the same `db` package than `db_test`. This is explicitly allowed by the google styleguide.

PR #228, opened 2025-04-10
fix: Fixes CA status setting when revoking certs

saltiyazan:
# Description

When revoking a leaf cert the status of the the issuer's cert should not be changed, but when a CA is revoked, its state should become "legacy"

# Checklist:

### DIFF
--- a/internal/db/db_certificate_authorities.go
+++ b/internal/db/db_certificate_authorities.go
@@ -518,12 +518,11 @@ func (db *Database) SignCertificateRequest(csrFilter CSRFilter, caFilter Certifi
 	return err
 }
 
-// RevokeCertificate revokes a certificate previously signed by a Notary CA by places the serial number of the certificate in its CRL.
+// RevokeCertificate revokes a certificate previously signed by a Notary CA by placing the serial number of the certificate in its issuer's CRL.
 func (db *Database) RevokeCertificate(filter CSRFilter) error {
 	oldRow, err := db.GetCertificateRequestAndChain(filter)
 	if err != nil {
-		log.Println(err)
-		return fmt.Errorf("%w: no certificate request found", ErrNotFound)
+		return err
 	}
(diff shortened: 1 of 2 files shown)


internal/db/db_certificate_authorities.go
```diff
@@ -535,13 +534,11 @@ func (db *Database) RevokeCertificate(filter CSRFilter) error {
 	}
 	issuerCert, err := db.GetCertificate(ByCertificatePEM(certChain[1]))
 	if err != nil {
-		log.Println(err)
-		return fmt.Errorf("%w: issuer certificate not found", ErrInternal)
+		return err
```
kayra1, line 537: why are we going back to this format? I thought we wanted to log the detailed errors and return an HTTP request friendly error?

internal/server/handlers_certificate_authorities_test.go
```diff
@@ -1628,8 +1628,8 @@ func TestCertificateRevocationListsEndToEnd(t *testing.T) {
 		if err != nil {
 			t.Fatalf("expected no error, got: %s", err)
 		}
-		if cas.Result[1].Status != "pending" {
-			t.Fatalf("expected revoked intermediate CA to have revoked status")
+		if cas.Result[1].Status != "legacy" {
+			t.Fatalf("expected revoked intermediate CA to have legacy status")
```
kayra1, line 1632: Can you also add a check for this change in line 1600 to make sure that the changes you've made work?

kayra1 (changes requested):

internal/db/db_certificate_authorities.go
saltiyazan: Here we are handling the errors of a db function, all of those should have already correctly handled the errors and logged them. We don't need to do that 2 times.

kayra1 (approved):

PR #226, opened 2025-04-09
feat: certificate authority management page

kayra1:
# Description

This PR introduces a new page to the frontend where the new CA API changes can be used. It includes all of the required features for CA management and signing except for revoking all sub-certificates.
 
image

image

image

image

# Checklist:

- [x] My code follows the [style guidelines](/CONTRIBUTING.md) of this project
- [x] I have performed a self-review of my own code
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that validate the behaviour of the software
- [x] I validated that new and existing unit tests pass locally with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #232, opened 2025-04-22
feat: add structured logging with zap

gruyaume:
# Description

We add structured logging, standardizing how and what Notary logs. This task is the implementation of the specification referenced below. Here we focus on system logs, audit logs will be taken care of separately.

Each log entry has the following content:
- Level
- Timestamp
- Caller
- Message

## Log Example

```
{"level":"info","ts":"2025-04-22T17:21:56.748-0400","caller":"server/middleware.go:98","msg":"Request: GET /certificate_requests 200 OK"}
{"level":"info","ts":"2025-04-22T17:21:58.316-0400","caller":"server/middleware.go:98","msg":"Request: GET /accounts 200 OK"}
{"level":"info","ts":"2025-04-22T17:22:11.896-0400","caller":"server/middleware.go:98","msg":"Request: POST /accounts 201 Created"}
{"level":"info","ts":"2025-04-22T17:22:11.899-0400","caller":"server/middleware.go:98","msg":"Request: GET /accounts 200 OK"}
```

## Config file Changes

We added the `logging` block:

### Example 1: Stdout logging
```yaml
logging:
  system:
    level: "debug"
    output: "stdout"
```

### Example 2: File logging

```yaml
logging:
  system:
    level: "debug"
    output: "notary.log"
```

## Reference
- https://betterstack.com/community/guides/logging/go/zap/
- https://docs.google.com/document/d/1KvSD6XLVvkrxKkb2qeqlj9utbw2buYndLzQPlszTH9s/edit?tab=t.0#heading=h.37fow4xa5j76

# Checklist:

### DIFF
--- a/cmd/notary/main.go
+++ b/cmd/notary/main.go
@@ -26,7 +26,7 @@ func main() {
 	}
 	logger, err := l.NewLogger(&conf.Logging)
 	if err != nil {
-		logger.Fatalf("Couldn't create logger: %s", err)
+		log.Fatalf("Couldn't create logger: %s", err)
 	}
 	srv, err := server.New(conf.Port, conf.Cert, conf.Key, conf.DBPath, conf.ExternalHostname, conf.PebbleNotificationsEnabled, logger)
 	if err != nil {
--- a/cmd/notary/main.go
+++ b/cmd/notary/main.go
@@ -24,11 +24,7 @@ func main() {
 	if err != nil {
 		log.Fatalf("Couldn't validate config file: %s", err)
 	}
-	l, err := logger.NewLogger(&logger.LoggerOpts{
-		System: logger.SystemLoggerOpts{
-			Level: "info",
-		},
-	})
+	l, err := logger.NewLogger(&conf.Logging)
 	if err != nil {
 		l.Fatalf("Couldn't create logger: %s", err)
 	}
--- a/cmd/notary/main.go
+++ b/cmd/notary/main.go
@@ -9,6 +9,7 @@ import (
 	"os/signal"
 
 	"github.com/canonical/notary/internal/config"
+	"github.com/canonical/notary/internal/logger"
 	"github.com/canonical/notary/internal/server"
 )
 
@@ -23,9 +24,17 @@ func main() {
 	if err != nil {
 		log.Fatalf("Couldn't validate config file: %s", err)
 	}
-	srv, err := server.New(conf.Port, conf.Cert, conf.Key, conf.DBPath, conf.ExternalHostname, conf.PebbleNotificationsEnabled)
+	l, err := logger.NewLogger(&logger.LoggerOpts{
+		System: logger.SystemLoggerOpts{
+			Level: "info",
+		},
+	})
 	if err != nil {
-		log.Fatalf("Couldn't create server: %s", err)
+		l.Fatalf("Couldn't create logger: %s", err)
+	}
+	srv, err := server.New(conf.Port, conf.Cert, conf.Key, conf.DBPath, conf.ExternalHostname, conf.PebbleNotificationsEnabled, l)
+	if err != nil {
+		l.Fatalf("Couldn't create server: %s", err)
 	}
 
 	idleConnsClosed := make(chan struct{})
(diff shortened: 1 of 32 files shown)


gruyaume:
@ghislainbourgeois  I assigned the review explicitly to you because I thought you may have opinions/ideas around logging. Let me know what you thing. Note that I'm only keeping it in draft in order to merge #226 first, since that one is related to our cycle and pulse objectives, but it is ready for review.

cmd/notary/main.go
```diff
@@ -9,6 +9,7 @@ import (
 	"os/signal"
 
 	"github.com/canonical/notary/internal/config"
+	"github.com/canonical/notary/internal/logger"
```
ghislainbourgeois, line 12: I think it looks a big weird to use `l` in this file for logging calls. We could alias the import instead, and at least the logging calls would look like the other files.

```suggestion
	l "github.com/canonical/notary/internal/logger"
```

internal/server/handlers_accounts.go
```diff
@@ -72,7 +71,7 @@ func ListAccounts(env *HandlerConfig) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		accounts, err := env.DB.ListUsers()
 		if err != nil {
-			writeError(w, http.StatusInternalServerError, "Internal Error")
+			writeError(w, http.StatusInternalServerError, "Internal Error", env.Logger)
```
ghislainbourgeois, line 74: I put the comment on this first line, but it is more general for the PR as a whole. We removed the logging calls of the specific errors when an error occurs and instead log them here, but we lose the specifics.

I would suggest adding an error parameter to `writeError` so that we can log those details. Otherwise, we would only log `Internal Error` here, with no details available to investigate the reason.

```suggestion
			writeError(w, http.StatusInternalServerError, "Internal Error", err, env.Logger)
```

The same comment would apply in most places, but particularly where we have an internal error.

ghislainbourgeois (changes requested):
In general this looks good, but I think we are losing a bit too much details in the logs with the current state of the PR. I have added a suggestion on how we could fix this.

cmd/notary/main.go
gruyaume: Ok sure, Done!

DanielArndt:
I love structured logging, and I'd like to give an entirely subjective opinion/suggestion on it: We should make all log messages static and only allow variables in the key-value pairs. This makes it a lot easier to process the logs if you're looking for something specific. Ensure that strings are never split across lines, and it will also be trivial to find the spot a log came from.

In our case, that means 

```
log.Fatalf("Couldn't validate config file: %s", err)
```

becomes

```
log.Fatalf("Couldn't validate config file", 
	"err", 
	err
)
```

(if I have my syntax correctly).

I am still going through the rest of the PR, but I wanted to throw this out there for feedback.

docs/reference/config_file.md
```diff
@@ -11,6 +11,13 @@ Start Notary with the `--config` flag to specify the path to the configuration f
 - `db_path` (string): Path to where the sqlite database should be stored. If the file does not exist Notary will attempt to create it.
 - `port` (integer): Port number on which Notary will listen for all incoming API and frontend connections.
 - `pebble_notifications` (boolean): Allow Notary to send pebble notices on certificate events (create, update, delete). Pebble needs to be running on the same system as Notary.
+- `logging` (object): Configuration for logging.
+  - `system` (object): Configuration for system logging.
+    - `level` (string): The level of logging. Options are `debug`, `info`, `warn`, `error`, and `fatal`.
+    - `output` (string): The output destination for logs. Options are `stdout`, and `file`.
+      - `stdout`: Logs will be printed to the console.
+      - `file`: Logs will be written to a file. This option requires the `path` parameter to be specified.
```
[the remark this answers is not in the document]
gruyaume: Right now that is not supported.

docs/reference/config_file.md
```diff
@@ -11,6 +11,13 @@ Start Notary with the `--config` flag to specify the path to the configuration f
...
+    - `path` (string, optional): The path to the log file. If not specified, logs will be printed to stdout. This option is only applicable when `output` is set to `file`.
```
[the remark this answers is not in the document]
gruyaume: Yes, we could have `output` be a string instead of an object and have users able to specify `stdout` or the file path. But that's not the spec that was approved. If we are confident that it's a better idea, I'm ok with moving to that. 

Reference:
- https://docs.google.com/document/d/1KvSD6XLVvkrxKkb2qeqlj9utbw2buYndLzQPlszTH9s/edit?tab=t.0

gruyaume:
> I love structured logging, and I'd like to give an entirely subjective opinion/suggestion on it: We should make all log messages static and only allow variables in the key-value pairs. This makes it a lot easier to process the logs if you're looking for something specific. Ensure that strings are never split across lines, and it will also be trivial to find the spot a log came from.

Right now we are using Zap's `SugaredLogger` which is a bit more convenient, less type safe, and slower than the `Logger` approach you are suggesting.

```go
logger.Fatalf("Couldn't create server: %s", err)  // SugaredLogger
logger.Fatal("Couldn't create server", zap.String("err", err.Error()))  // Logger
```

Reading more about it pushes me to use the `Logger` approach, it'll be safer, which would be appropriate for Notary.

ghislainbourgeois (approved):
LGTM

internal/server/response.go
```diff
@@ -34,22 +35,22 @@ func writeResponse(w http.ResponseWriter, v any, status int) error {
 }
 
 // writeError is a helper function that logs any error and writes it back as an http response
-func writeError(w http.ResponseWriter, status int, message string) {
+func writeError(w http.ResponseWriter, status int, message string, err error, logger *zap.Logger) {
```
saltiyazan, line 38: The docstring of this function can benefit from some clarification now that it got new parameters.
The error itself is not written to the http response (and it shouldn't be) only the message is.

gruyaume: Ok I adapted the docstring

saltiyazan (approved):

PR #249, opened 2025-06-11
feat: replace CA status with `enabled` field

### LINKED ISSUE #225: The status of the CA is not updated on expiry
#### Describe the bug
The CA remains "Active" even after it is expired.

#### To Reproduce

1. Run Notary
2. Create a short living CA
3. List active CAs
4. Wait until the CA expires
5. List active CAs again, the CA is still there, its status has not changed.

#### Expected behavior
I'm not sure how, but the CA should probably go into the "expired" state

gruyaume:
# Description

Notary maintained a CA status that could be either "active", "pending", "legacy", or "expired". Here we remove all these options and replace them with a simpler `enabled` boolean flag. Whether the CA is pending, or expired can be inferred from other values. For instance, `expired` can be inferred from the certificate content comparing the cert valid_before time with now. And `pending` can be inferred from whether there is a certificate associated with the certificate request. The only un-inferrable field is `enabled` which represents the user's intent as to whether this CA can be used to sign certificates or not. This field is completely independent from the certificate content. A certificate can be expired or valid and be enabled or disabled, it purely represents user's intent. By the fact that `enabled` is completely independent from other fields, we don't need a loop to automatically change it. 

Fixes #225 

# Checklist:


internal/server/server.go
```diff
@@ -101,3 +120,45 @@ func New(port int, cert []byte, key []byte, dbPath string, externalHostname stri
 
 	return s, nil
 }
+
+// reconcileCAStatus checks the status of all CAs in the database and updates their status
+// if necessary.
+func ReconcileCAStatus(dbClient *db.Database, logger *zap.Logger) error {
```
ghislainbourgeois, line 126: I understand that you had to make this public to be able to test it, but we should not have to. We should make our `_test.go` files part of the same package, but I tried it locally and at least for this package, it is a relatively big change to the test.

I think we should refactor this in the near future in another PR.

internal/server/server.go
```diff
@@ -101,3 +120,45 @@ func New(port int, cert []byte, key []byte, dbPath string, externalHostname stri
...
+	if err != nil {
+		return fmt.Errorf("failed to list certificate authorities: %w", err)
+	}
+
+	for _, ca := range certificateAuthorities {
+		caDenorm, err := dbClient.GetDenormalizedCertificateAuthority(
+			db.ByCertificateAuthorityDenormalizedID(ca.CertificateAuthorityID),
+		)
+		if err != nil {
+			return fmt.Errorf("failed to get denormalized certificate authority: %w", err)
```
ghislainbourgeois, line 137: In this case, I am not sure we should return the error, but probably only log it and continue. Otherwise, a broken CA that appears early in the list would block the whole process.

gruyaume: done!

internal/server/server.go
```diff
@@ -101,3 +120,45 @@ func New(port int, cert []byte, key []byte, dbPath string, externalHostname stri
...
+		cert, err := x509.ParseCertificate(block.Bytes)
+		if err != nil {
+			logger.Warn("failed to parse certificate", zap.Int64("ca_id", ca.CertificateAuthorityID), zap.Error(err))
+			continue
+		}
+
+		if time.Now().After(cert.NotAfter) && ca.Status != db.CAExpired {
+			err = dbClient.UpdateCertificateAuthorityStatus(db.ByCertificateAuthorityID(ca.CertificateAuthorityID), db.CAExpired)
+			if err != nil {
+				return fmt.Errorf("failed to update CA status to expired: %w", err)
```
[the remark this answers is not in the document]
gruyaume: done!

internal/server/server.go
gruyaume: I could still write the test under the same package as code here and make the function private if we think that's the right path. The only downside for now is that we'd have a mix of tests being in the same package or in a _test one.

kayra1 (changes requested):
Most of it looks good, except there are a bunch of things that Go and sqlite can do that can be utilized here:

1) SQLite treats the strings "true" and "false" as aliases for 1 and 0 respectively:
```	
INSERT INTO certificate_authorities (crl, enabled, private_key_id, csr_id, certificate_id) 
VALUES (<crl>, true, 1, 1, null)
```
works because SQlite will treat "true" as an alias for 1.

2) SQLair casts 1 and 0 into the correct boolean type when scanning.
```
SELECT &CertificateAuthority.* FROM certificate_authorities WHERE certificate_authority_id==$CertificateAuthority.certificate_authority_id
```
will result in correctly filling out the enabled field:
```
CertificateAuthority{
		CertificateAuthorityID: 0,
		CertificateID:          0,
		Enabled:                false,
		CRL: "string",
		PrivateKeyID: 0,
		CSRID: 0,
	}

```
This means we can just mark the enabled field as `bool` in Go, and sqlair and go will correctly serialize and deserialize the data to and from the database. the table definition for the enabled row should still remain as INTEGER.

gruyaume:
> Most of it looks good, except there are a bunch of things that Go and sqlite can do that can be utilized here:
> 
> 1. SQLite treats the strings "true" and "false" as aliases for 1 and 0 respectively:
> 
> ```
> INSERT INTO certificate_authorities (crl, enabled, private_key_id, csr_id, certificate_id) 
> VALUES (<crl>, true, 1, 1, null)
> ```
> 
> works because SQlite will treat "true" as an alias for 1.
> 
> 2. SQLair casts 1 and 0 into the correct boolean type when scanning.
> 
> ```
> SELECT &CertificateAuthority.* FROM certificate_authorities WHERE certificate_authority_id==$CertificateAuthority.certificate_authority_id
> ```
> 
> will result in correctly filling out the enabled field:
> 
> ```
> CertificateAuthority{
> 		CertificateAuthorityID: 0,
> 		CertificateID:          0,
> 		Enabled:                false,
> 		CRL: "string",
> 		PrivateKeyID: 0,
> 		CSRID: 0,
> 	}
> ```
> 
> This means we can just mark the enabled field as `bool` in Go, and sqlair and go will correctly serialize and deserialize the data to and from the database. the table definition for the enabled row should still remain as INTEGER.

Thank you, I wasn't aware. This change was added in 06ebdde58bf637ce51f4ef20fb4afceb101f9eb1

PR #250, opened 2025-06-11
feat: map certificate requests to users

gruyaume:
# Description

Here we map certificate requests to users, making it clear to certificate administrators who requested which certificate. This feature is valuable on its own and as part of adding authorization support to Notary. This includes changes at the database, API, and UI levels.

## API

We added `username` as a return attribute when retrieving certificate requests `GET https://127.0.0.1:3001/api/v1/certificate_requests/1`

```
{
    "result": {
        "id": 1,
        "csr": "-----BEGIN CERTIFICATE REQUEST-----\nMIIC6TCCAdECAQAwcjELMAkGA1UEBhMCQ0ExCzAJBgNVBAgMAlFDMRIwEAYDVQQH\nDAlNb250csOpYWwxFzAVBgNVBAoMDkNhbm9uaWNhbCBMdGQuMRcwFQYDVQQLDA5D\nYW5vbmljYWwgTHRkLjEQMA4GA1UEAwwHZGRkLmNvbTCCASIwDQYJKoZIhvcNAQEB\nBQADggEPADCCAQoCggEBALb0EezLypyljI1c+prNLIqyOMGJ8U5M85HutF2vy5vs\nYH9WFWkYbXto7M18BKLRHJ1AA9WFeNFGyYwsu+2RESYTCUrTGh2RFlpqvVoanTMx\n57/iDA0Y7FHZByy1HaMyHAui7GSNVutuNUAmwyoVsFcFAVXMh+VSy/FuCmy6Fco1\nJLpu/GVae5Hfx3VV0suktAdEmaV+eMfuxCgUGawPmYIVHHs4FnayYv7SOFvF67fQ\nG8w+NH0jPnIb3AxDsFSUQeTbmhCfHN9pxeDggnpiY1QOj8lxzvVG9xfzQ93cmtjb\ngCT3frs4WtXNPF2H/Kz5ID7dDwslipfWCgvqTUWlmvcCAwEAAaAyMDAGCSqGSIb3\nDQEJDjEjMCEwHwYDVR0RBBgwFoIHZGRkLmNvbYILd3d3LmRkZC5jb20wDQYJKoZI\nhvcNAQELBQADggEBAFhbU7bayznci/TT5R1lCAbrcqYSKdndx00r+gI+NVI2p/mY\n9sNQD3P7loAP1Qj8Gw1QS4Qy1eHzErc1G+s3ujPdasB6zNlP7QVT8uNfHJpQoxNc\nG+Ga1CWvwEC3O0yF9jfqrIyDQW4CNhRVAXwVeiX4hM3qTwGAwV+o/EjWhMENvcuE\nItoO9PfHxNjJCvg7GzoKJapd61yXKr5mf14FUN2FO3oz71xgkndjVKRszX4EETOl\nNYgiZXAeRRWaqQ2a9DIm8mwRvp21f6cug58MA+vm5bYjeCDqhvE3UfA+T/WmauOh\nLwBjBCi42LNHDrFcUQkaesQQ4Gr3wjCRyjhjOf0=\n-----END CERTIFICATE REQUEST-----",
        "certificate_chain": "",
        "status": "Outstanding",
        "username": "gruyaume"
    }
}
```

## Screenshot

image

## Reference
- https://docs.google.com/document/d/1cjOmWbiRkswyzcILrRUt16NMw0INs3ywj7M32i3VlCU/edit?tab=t.0

# Checklist:

### DIFF
--- a/internal/db/db_certificate_authorities_test.go
+++ b/internal/db/db_certificate_authorities_test.go
@@ -25,7 +25,7 @@ func TestRootCertificateAuthorityEndToEnd(t *testing.T) {
 		t.Fatalf("CA found when no CA's should be available")
 	}
 
-	caID, err := database.CreateCertificateAuthority(RootCACSR, RootCAPrivateKey, RootCACRL, RootCACertificate+"\n"+RootCACertificate)
+	caID, err := database.CreateCertificateAuthority(RootCACSR, RootCAPrivateKey, RootCACRL, RootCACertificate+"\n"+RootCACertificate, 0)
 	if err != nil {
 		t.Fatalf("Couldn't create certificate authority: %s", err)
 	}
@@ -103,7 +103,7 @@ func TestIntermediateCertificateAuthorityEndToEnd(t *testing.T) {
 		t.Fatalf("CA found when no CA's should be available")
 	}
 
-	caID, err := database.CreateCertificateAuthority(IntermediateCACSR, IntermediateCAPrivateKey, "", "")
+	caID, err := database.CreateCertificateAuthority(IntermediateCACSR, IntermediateCAPrivateKey, "", "", 0)
 	if err != nil {
 		t.Fatalf("Couldn't create certificate authority: %s", err)
 	}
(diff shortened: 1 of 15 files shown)


internal/db/sql_stmts.go
saltiyazan, line 1: I know this PR followed what we already do in the codebase so this comment can considered as a more general one but I just noticed that currently, our database schema does not enforce any foreign key constraints. This means that it is possible to insert records that reference non-existent rows in other tables. For example, you can create a certificate request with a user ID that does not actually exist in the users table, or a certificate with an issuer ID that does not exist in the certificates table.
Although the API is designed to pass valid IDs (such as extracting the user ID from a JWT, which should correspond to an existing user), there is no guarantee at the database level or in the application logic that these references are valid. As a result, there is a risk of introducing data integrity issues if mistakes are made elsewhere in the code.
This limitation applies to all statements that reference external data, such as creating certificate requests, certificates, or certificate authorities.

What do you think @kayra1 ?

Should we have such constraints?

internal/db/db_certificate_authorities_test.go
```diff
@@ -25,7 +25,7 @@ func TestRootCertificateAuthorityEndToEnd(t *testing.T) {
 		t.Fatalf("CA found when no CA's should be available")
 	}
 
-	caID, err := database.CreateCertificateAuthority(RootCACSR, RootCAPrivateKey, RootCACRL, RootCACertificate+"\n"+RootCACertificate)
+	caID, err := database.CreateCertificateAuthority(RootCACSR, RootCAPrivateKey, RootCACRL, RootCACertificate+"\n"+RootCACertificate, 0)
```
saltiyazan, line 28: Does a user with the ID `0` exist? Please see related comment, this might not be an issue.

saltiyazan (commented):
The PR looks good, I tested it too.
I just would like to take the opportunity to discuss foreign key constraints in the schema.

internal/db/db_certificate_authorities_test.go
gruyaume: This was fixed in b15ae23871a1735449ef189532fccc6b646101f5

internal/db/db_csrs_test.go
saltiyazan, line 1: Should we add a test that validates the CSR will not be created if we provide an ID of a user that doesn't exist?

internal/db/sql_stmts.go
saltiyazan, line 1: Should we think of what must happen when a user is delete? do we delete the csr it had created? I'm not saying we should, I just want to make sure we think about it, feel free to resolve and merge if you don't think we should do something about it in this PR.

saltiyazan (approved):
Looks good.
I requested one extra test case. But feel free to resolve and merge when you add it.

internal/db/sql_stmts.go
gruyaume: Yes, I have thought of this.

We won't delete the CSR or certificate is the user is deleted. Imagine if an employee made a request, got a certificate, then leaves the company, the certificate is still relevant, regardless of that employee's employment status. So we simply make the user_id field "null" and respond "unknown" in API requests.

kayra1 (commented):
The reason there are currently no foreign keys in the SQL statements is because they are not supported by a bunch of valid database options, and that decision was made by me when we valued using only the shared subset of SQL features in case we wanted to move off of SQLite in the future. 

Also, [sqlite claims](https://sqlite.org/foreignkeys.html) you will need to run `PRAGMA foreign_keys = ON;` alongside the first table definitions to enable referential integrity checks like ON DELETE, so a test for that would be nice.

internal/db/db_csrs_test.go
gruyaume: Agreed, we now validate this part of the `TestCertificateRequestUserMappingEndToEnd` database unit tests test. Ref: 38e9a089917c0bb7119210128882d54ca281effb

saltiyazan (approved):

PR #253, opened 2025-06-13
feat: Adds encryption backend config and support for PKCS11 backend

saltiyazan:
# Description

- Adds the encryption backend configs (`vault`, `pkcs11`, `none`)
- Adds support for `pkcs11` and `none`
- Encrypts and decrypts the encryption key using the configured backend
- Fixes tests
- Updates docs

**Note**: This implementation doesn't allow migrating from one encryption backend type to another, the user will need to use the same encryption backend they configured at the time of initialisation. Please comment on this matter if you think we should address it in the context of this change (if we want to allow such migration, because it might not be simple if it isn't simply from none to a an actual backend)

To run this with an HSM:
- Provide required configs. Make sure your HSM is connected, the connector is running. To create the key you can use a command like `generate symmetric 0 0x1234 aes256-cbc-key 1 encrypt-cbc,decrypt-cbc aes256`. 
- ` export YUBIHSM_PKCS11_CONF="/Users/yazan/repos/work/notary/cmd/notary/yubihsm_pkcs11.conf"`, example of the config [here](https://developers.yubico.com/yubihsm-shell/yubihsm-pkcs11.html).

# Checklist:


internal/config/config.go
```diff
@@ -143,5 +174,60 @@ func Validate(filePath string) (Config, error) {
...
 	config.Logging.System.Level = LoggingLevel(c.Logging.System.Level)
 	config.Logging.System.Output = c.Logging.System.Output
+	config.EncryptionBackend, err = createEncryptionBackend(c.EncryptionBackend)
+	if err != nil {
+		return Config{}, fmt.Errorf("failed to create encryption backend: %w", err)
+	}
 	return config, nil
 }
+
+// createEncryptionBackend creates a SecretBackend based on the type specified
```
gruyaume, line 184: It feels wrong for the config code to create an encryption backend. The config file and code is really only meant for fetching and validating config options. I think this function and calling it belongs elsewhere. In other words, `internal/config` should probably not import other `internal/...` packages.

internal/encryption/pkcs11.go
```diff
@@ -0,0 +1,194 @@
...
+)
+
+// PKCS11Backend implements EncryptionBackend using PKCS11 protocol for HSMs.
+type PKCS11Backend struct {
+	ctx   *pkcs11.Ctx
+	pin   string
+	keyID uint16
+}
+
+const ivSize = 16
```
gruyaume, line 17: Since it's in bytes, I would recommend adding a `// bytes` next to the value to make it clear to developers what we are talking about. Or even add bytes to the variable name.

internal/encryption/pkcs11.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+// Encrypt encrypts the plaintext using AES-CBC algorithm.
+func (h *PKCS11Backend) Encrypt(data []byte) ([]byte, error) {
+	session, key, err := h.connectToBackend()
+	if err != nil {
+		return nil, fmt.Errorf("failed to connect to backend: %w", err)
+	}
+	defer func() {
+		if cleanupErr := h.cleanupSession(session); cleanupErr != nil {
```
gruyaume, line 40: Can't we use the logger instead of `fmt.Printf`? The same comment may apply elsewhere

internal/encryption/pkcs11.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	slots, err := h.ctx.GetSlotList(true)
+	if err != nil || len(slots) == 0 {
+		if cleanupErr := h.cleanupInitialization(); cleanupErr != nil {
+			fmt.Printf("Warning: cleanup error after GetSlotList failure: %v\n", cleanupErr)
+		}
+		return 0, 0, err
+	}
+	slot := slots[0]
```
gruyaume, line 156: I don't know all that much about hsm's yet but is there good reason to assume the first slot?

internal/encryption/pkcs11.go
```diff
@@ -0,0 +1,194 @@
...
+
+	session, err := h.ctx.OpenSession(slot, pkcs11.CKF_SERIAL_SESSION|pkcs11.CKF_RW_SESSION)
+	if err != nil {
+		if cleanupErr := h.cleanupInitialization(); cleanupErr != nil {
+			fmt.Printf("Warning: cleanup error after OpenSession failure: %v\n", cleanupErr)
```
gruyaume, line 161: Same comment, we have a logger for logging, avoid using `fmt.Printf`

internal/encryption/pkcs11.go
gruyaume, line 1: We seem to be lacking tests for this package. With interfaces you could mock the return of hsm's or Vault and make it possible to test the logic that we have here.

gruyaume:
> **Note**: This implementation doesn't allow migrating from one encryption backend type to another, the user will need to use the same encryption backend they configured at the time of initialisation. Please comment on this matter if you think we should address it in the context of this change (if we want to allow such migration, because it might not be simple if it isn't simply from none to a an actual backend)

While I do think this is a reasonable constraints, we should document this to users. Also, what happens when a user tries to change the backend encryption, are we clearly letting the user know of what's the issue?

kayra1:
> While I do think this is a reasonable constraints, we should document this to users. Also, what happens when a user tries to change the backend encryption, are we clearly letting the user know of what's the issue?

This is actually something that we have designed in mind to be able to do. It is still going to be (relatively) complicated, so I think it's fine to not have it here for now. Please do or let me know and I can create a ticket tracking this feature on JIRA @saltiyazan

docs/reference/config_file.md
```diff
@@ -15,11 +15,22 @@ Start Notary with the `--config` flag to specify the path to the configuration f
   - `system` (object): Configuration for system logging.
     - `level` (string): The level of logging. Options are `debug`, `info`, `warn`, `error`, and `fatal`.
     - `output` (string): The output destination for logs. Options are `stdout`, `stderr`, or a file path.
+- `encryption_backend` (object): Configuration for the encryption backend
+  - `type` (string): Type of the encryption backend, currently supported backends are `vault`, `pkcs11` and `none` for no encryption backend.
+  - `endpoint` (string): URL endpoint for the Vault server when using the Vault backend.
```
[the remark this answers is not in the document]
kayra1: I think my preference would be:
```
encryption_backend:
  vault:
    endpoint: "https://vault.com"
    role_secret_id: "12345"
    ...
```
We can just use the top level key as the type directly.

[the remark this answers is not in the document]
kayra1: Me, Yazan and @DanielArndt spoke on the multiple encryption backend options and we came up with the idea that we should have multiple backends that should work simultaneously, but this should be done at a future PR, and we should define the features in a spec together.

internal/encryption/pkcs11.go
saltiyazan: Replaced with logger.

internal/encryption/pkcs11.go
saltiyazan: Good point. This would have almost always worked for yubihsm2, but it nothing guarantees it always works for any hsm. hsms can have multiple slots, each slot will have a different token and our object (encryption key) might not be in the first slot. Since the order of the slots is not always necessarily the same, and slots are not configured by user (as far as I understood) I took an approach of trying to find the key in each slot from the list (using the key id and the pin). In most cases we will see one slot and try once.

internal/encryption/pkcs11.go
saltiyazan: :white_check_mark:

internal/encryption/encryption_backend.go
gruyaume, line 1: I am somewhat confused about the various encrypt/decrypt functionalities in Notary. Under `internal/encryption` we have 3 different files providing `Encrypt` and `Decrypt` endpoints. I understand the No encryption vs PKCS11, but the third one under `internal/encryption.go` is confusing here. That one is different right? I think there's something we could do in reorganizing the directories. Maybe:

```
internal/
  encryption_backend/
    encryption_backend.go // (contains only the interface)
    no_encryption.go
    pkcs11.go
    ..other backends would belong here
  encryption/
    encryption.go
```

We should also explain how we do encryption. It's not directly obvious why we need the local encrypt/decrypt even though we have HSM support.

internal/encryption/encryption_backend.go
```diff
@@ -0,0 +1,23 @@
...
+
+// Decrypt returns the ciphertext as is for NoEncryptionBackend.
+func (n NoEncryptionBackend) Decrypt(ciphertext []byte) ([]byte, error) {
+	return ciphertext, nil
+}
+
+// Encrypt returns the plaintext as is for NoEncryptionBackend.
+func (n NoEncryptionBackend) Encrypt(plaintext []byte) ([]byte, error) {
+	return plaintext, nil
+}
```
gruyaume, line 23: As mentioned in my comment above, the "no encryption" content can belong in its own file, i.e. 1 file per backend.

internal/encryption/pkcs11.go
```diff
@@ -0,0 +1,271 @@
...
+			h.logger.Error("Error during cleanup of PKCS11 resources while connecting to backend", zap.Error(cleanupErr))
+		}
+		return 0, 0, err
+	}
+
+	var session pkcs11.SessionHandle
+	var keyHandle pkcs11.ObjectHandle
+	var lastErr error
+
+	for _, slot := range slots {
```
gruyaume, line 225: Which piece would fail if we try to use the wrong slot? Is it just the `OpenSession` call? If yes, I would recommend writing a simple `openSessionOnAvailableSlot` (or find a better name) helper that has a smaller loop and only return the session and there'd be no loop in this function.

internal/encryption/pkcs11_test.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+	return plaintext, nil
+}
+
+func TestPKCS11Backend_EncryptDecryptEndToEnd(t *testing.T) {
+	tests := []struct {
+		name      string
+		message   []byte
+		wantErr   bool
+		errorCase string
```
gruyaume, line 110: This field is unclear to me, why do we capture "short" independently, and have a switch case for it. Does this really belong in a table test if we need switch cases?

CONTRIBUTING.md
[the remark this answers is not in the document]
saltiyazan: Added

cmd/notary/main.go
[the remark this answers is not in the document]
saltiyazan: I added minimal logging that will tel the user when the backend has been configured, whether a new encryption key has been created and encrypted or the one from the db is being used.

docs/reference/config_file.md
```diff
@@ -15,17 +15,44 @@ Start Notary with the `--config` flag to specify the path to the configuration f
   - `system` (object): Configuration for system logging.
     - `level` (string): The level of logging. Options are `debug`, `info`, `warn`, `error`, and `fatal`.
     - `output` (string): The output destination for logs. Options are `stdout`, `stderr`, or a file path.
+- `encryption_backend` (string or object): Configuration for the encryption backend. Can be either "none" for no encryption, or an object containing named backend configurations.
```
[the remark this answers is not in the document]
saltiyazan: I think something like `encryption_backend: none` is more readable and explicit than something like `encryption_backend:`
Yes, it adds a bit of complexity to the unmarshalling logic but I think it makes for a better user experience.

docs/reference/config_file.md
```diff
@@ -15,17 +15,44 @@ Start Notary with the `--config` flag to specify the path to the configuration f
...
+  - `backend_name` (object): User-defined name for the encryption backend (e.g., "yubihsm", "hsm1").
+    - `pkcs11` (object): Configuration for PKCS#11 backend.
+      - `lib_path` (string): Path to the PKCS#11 library needed to communicate with the backend.
+      - `pin` (string): PIN for authenticating with the PKCS#11 device.
+      - `key_id` (integer): ID of the key to use on the PKCS#11 device.
 
-## Example
+## Examples
 
+### With PKCS#11 Backend
```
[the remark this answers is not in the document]
saltiyazan: :white_check_mark:

docs/reference/config_file.md
```diff
@@ -15,17 +15,44 @@ Start Notary with the `--config` flag to specify the path to the configuration f
...
+    output: "/var/lib/notary/logs/notary.log"
+encryption_backend:
+  yubihsm:
+    pkcs11:
+      lib_path: "/path/to/yubihsm_pkcs11.so"
+      pin: "0001password"
+      key_id: 0x1234
+```
+
+### With No Encryption
```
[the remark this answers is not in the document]
saltiyazan: :white_check_mark:

docs/index.md
```diff
@@ -12,6 +12,10 @@ We designed Notary for Enterprise environments where users need to securely mana
 - **Simple UI**: A simple and intuitive web interface for managing certificates.
 - **Extensive HTTP API**: Accomplish all the tasks you can do in the UI via the HTTP API.
 - **Metrics**: Monitor the state of your certificates and the health of your Notary instance with Prometheus metrics.
+- **Encryption at Rest**: Sensitive data like private keys is encrypted in the database.
```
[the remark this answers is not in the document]
saltiyazan: :white_check_mark:

internal/config/config.go
```diff
@@ -135,6 +207,44 @@ func Validate(filePath string) (Config, error) {
...
+			}
+			if selectedBackend.PKCS11.KeyID == nil {
+				return Config{}, fmt.Errorf("key ID is missing")
+			}
+			backendConfig = BackendConfig{
+				Type:   PKCS11,
+				PKCS11: selectedBackend.PKCS11,
+			}
+		default:
+			return Config{}, fmt.Errorf("unknown backend type: invalid")
```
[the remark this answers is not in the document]
saltiyazan: :white_check_mark:

internal/encryption/pkcs11.go
```diff
@@ -0,0 +1,271 @@
...
+import (
+	"crypto/rand"
+	"fmt"
+
+	"github.com/miekg/pkcs11"
+	"go.uber.org/zap"
+)
+
+// PKCS11Context defines the interface for PKCS11 operations needed by our backend
+type PKCS11Context interface {
```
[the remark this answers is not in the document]
saltiyazan: :white_check_mark:

internal/encryption/pkcs11.go
```diff
@@ -0,0 +1,271 @@
...
+	FindObjects(pkcs11.SessionHandle, int) ([]pkcs11.ObjectHandle, bool, error)
+	FindObjectsFinal(pkcs11.SessionHandle) error
+	EncryptInit(pkcs11.SessionHandle, []*pkcs11.Mechanism, pkcs11.ObjectHandle) error
+	Encrypt(pkcs11.SessionHandle, []byte) ([]byte, error)
+	DecryptInit(pkcs11.SessionHandle, []*pkcs11.Mechanism, pkcs11.ObjectHandle) error
+	Decrypt(pkcs11.SessionHandle, []byte) ([]byte, error)
+}
+
+// pkcs11ContextWrapper implements PKCS11Context using the real pkcs11.Ctx
+type pkcs11ContextWrapper struct {
```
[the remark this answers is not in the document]
saltiyazan: :white_check_mark:

internal/encryption/pkcs11_test.go
```diff
@@ -0,0 +1,187 @@
...
+	"crypto/cipher"
+	"crypto/rand"
+	"fmt"
+	"testing"
+
+	"github.com/miekg/pkcs11"
+	"go.uber.org/zap"
+)
+
+type mockPKCS11Context struct {
```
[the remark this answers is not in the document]
saltiyazan: :white_check_mark:

internal/encryption/encryption_backend.go
saltiyazan: I moved the encryption backend stuff to their own dir.
But for clarity we encrypt data in Notary using a key managed by Notary and we encrypt that key using a key managed by the encryption backend if any is configured.

internal/encryption/pkcs11.go
saltiyazan: I can't give a definitive answer here, I couldn't find in the documents whether it is possible or not to login into different slots using the same pin. But if it was possible this will fail at the findkey.
I reorganized that bit of the code, I think it looks cleaner now.

docs/reference/config_file.md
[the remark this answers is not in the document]
gruyaume: What I'm proposing here is to simply not require the `encryption_backend` key when users don't use an encryption backend.

docs/index.md
```diff
@@ -12,6 +12,10 @@ We designed Notary for Enterprise environments where users need to securely mana
...
+- **Encryption Backend Support**: To increase security an external encryption backend can be configured to encrypt Notary's encryption key.
+    **Note**: Once Notary is initialized it must continue using the encryption backend configured at the time of initialization, at the moment there is no way to switch backends.
```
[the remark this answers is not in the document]
saltiyazan: Okay, I get your point, fixed.

docs/reference/config_file.md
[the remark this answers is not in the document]
saltiyazan: I still prefer the explicit approach but I don't know if there is a best practice here.

Any opinions @kayra1 ?

cmd/notary/main.go
```diff
@@ -53,3 +59,20 @@ func main() {
...
 }
+
+func createEncryptionBackend(backendConfig config.BackendConfig, logger *zap.Logger) (encryption_backend.EncryptionBackend, error) {
+	switch backendConfig.Type {
+	case config.PKCS11:
+		backend, err := encryption_backend.NewPKCS11Backend(backendConfig.PKCS11.LibPath, backendConfig.PKCS11.Pin, *backendConfig.PKCS11.KeyID, logger)
+		if err != nil {
+			return nil, err
+		}
+		logger.Info("PKCS11 backend created")
```
gruyaume, line 70: Sorry for being pedantic here, but we don't really "create" a backend, we connect to it. I'm pushing on these logs as they are the only way we currently inform users that they are using an HSM.

gruyaume (changes requested):

README.md
```diff
@@ -14,6 +14,8 @@ We designed Notary for Enterprise environments where users need to securely mana
 - **Simple UI**: A simple and intuitive web interface for managing certificates.
 - **Extensive HTTP API**: Accomplish all the tasks you can do in the UI via the HTTP API.
 - **Metrics**: Monitor the state of your certificates and the health of your Notary instance with Prometheus metrics.
+- **Encryption at Rest**: Sensitive data like private keys is encrypted in the database.
+- **Encryption Backend Support**: To increase security an external encryption backend can be configured to encrypt Notary's encryption key.
```
gruyaume, line 18: Please use the same content as in the docs index (i.e. only the point about hsm)

internal/encryption_backend/pkcs11.go
```diff
@@ -0,0 +1,274 @@
...
+func (p *realPKCS11Provider) DecryptInit(sh pkcs11.SessionHandle, m []*pkcs11.Mechanism, o pkcs11.ObjectHandle) error {
+	return p.ctx.DecryptInit(sh, m, o)
+}
+func (p *realPKCS11Provider) Decrypt(sh pkcs11.SessionHandle, message []byte) ([]byte, error) {
+	return p.ctx.Decrypt(sh, message)
+}
+
+// PKCS11Backend implements EncryptionBackend using the PKCS11 protocol for HSMs.
+type PKCS11Backend struct {
+	ctx    PKCS11Provider
```
gruyaume, line 73:
```suggestion
	pkcs11Provider    PKCS11Provider
```

Generally I'd avoid using `ctx` or `context` as variables, those are overly generic and encourage us to put anything in there. Also, those are generally used for `ctx context.Context` which we'll likely introduce in the future if/when we add tracing support.

gruyaume (changes requested):

docs/how-to/hsm.md
```diff
@@ -0,0 +1,29 @@
...
+4. Provide the ID of the symmetric encryption key that will be used
+
+```yaml
+encryption_backend:
+  yubihsm2-backend: # name of the backend
+    pkcs11:
+      lib_path: "/usr/lib/x86_64-linux-gnu/pkcs11/yubihsm_pkcs11.so"
+      pin: "0001password"
+      key_id: 0x1234
+```
```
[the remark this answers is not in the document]
gruyaume: What do you mean by " we can't specify a config file for the snap"? The config file is available after installing the snap under `/var/snap/notary/common/notary.yaml` and that's where users can change it prior to starting notary (`sudo snap start notary.notaryd`)

cmd/notary/main.go
```diff
@@ -29,7 +31,11 @@ func main() {
 	if err != nil {
 		log.Fatalf("Couldn't create logger: %s", err)
 	}
-	srv, err := server.New(conf.Port, conf.Cert, conf.Key, conf.DBPath, conf.ExternalHostname, conf.PebbleNotificationsEnabled, logger)
+	encryptionBackend, err := createEncryptionBackend(conf.EncryptionBackend, logger)
```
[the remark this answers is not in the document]
saltiyazan: There was a discussion over this with @DanielArndt that this doesn't belong to the database.
Even it is only used there. Is there any issue with creating this here? or is there any added value in moving it to the db?

docs/how-to/hsm.md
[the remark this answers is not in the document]
saltiyazan: I made them 2 points and the hsm related config is under 2.

README.md
saltiyazan: :white_check_mark:

internal/encryption_backend/pkcs11.go
saltiyazan: Renamed

docs/how-to/hsm.md
```diff
@@ -0,0 +1,45 @@
+# Use an HSM as Notary's Encryption Backend
+
+In this guide we show the required configuration to use a Hardware Security Module (HSM) as an encryption backend for Notary.
```
gruyaume, line 3:
```suggestion
In this guide we walk you through the required steps to configure and use a Hardware Security Module (HSM) as an encryption backend for Notary.
```

docs/how-to/hsm.md
```diff
@@ -0,0 +1,45 @@
...
+
+**Note**: 
+- Once Notary is initialized it must continue using the encryption backend configured at the time of initialization, at the moment there is no way to switch backends.
+- This has been tested with YubiHSM2, while this should work with any HSM that supports the PKCS11 protocol please follow any further instruction from the specific HSM vendor.
+
+## Hardware Security Modules (HSMs)
```
gruyaume, line 9:
```suggestion
```

docs/how-to/hsm.md
```diff
@@ -0,0 +1,45 @@
...
+
+### Prerequisites:
+- An HSM that supports the PKCS11 protocol
+- AES256 symmetric key created on the HSM with capabilities to encrypt and decrypt using the AES-CBC algorithm
+- Access to the HSM's driver/interface library file (.so or .dylib file) installed with the HSM's SDK
+- The HSM's connector up and running
+
+### Configuration
```
gruyaume, line 17:
```suggestion
### 1. Add your HSM information to Notary's configuration file
```

docs/how-to/hsm.md
```diff
@@ -0,0 +1,45 @@
...
+```yaml
+encryption_backend:
+  yubihsm2-backend: # name of the backend
+    pkcs11:
+      lib_path: "/usr/lib/x86_64-linux-gnu/pkcs11/yubihsm_pkcs11.so"
+      pin: "0001password"
+      aes_encryption_key_id: 0x1234
+```
+
+### Start Notary
```
gruyaume, line 34:
```suggestion
### 2. Start Notary
```

docs/how-to/hsm.md
```diff
@@ -0,0 +1,45 @@
...
+## Hardware Security Modules (HSMs)
+
+### Prerequisites:
+- An HSM that supports the PKCS11 protocol
+- AES256 symmetric key created on the HSM with capabilities to encrypt and decrypt using the AES-CBC algorithm
+- Access to the HSM's driver/interface library file (.so or .dylib file) installed with the HSM's SDK
+- The HSM's connector up and running
+
+### Configuration
+If you are using the snap you can modify the config under `/var/snap/notary/common/notary.yaml`
```
gruyaume, line 18: I think this line belongs in the config file reference, we don't want to add such a line in every how-to guide that relates to managing the config file.

docs/how-to/hsm.md
```diff
@@ -0,0 +1,45 @@
...
+      lib_path: "/usr/lib/x86_64-linux-gnu/pkcs11/yubihsm_pkcs11.so"
+      pin: "0001password"
+      aes_encryption_key_id: 0x1234
+```
+
+### Start Notary
+If you are using the snap start the service with the following command:
+```shell
+sudo snap start notary.notaryd
+```
```
gruyaume, line 38:
```suggestion
Start Notary:
```shell
sudo snap start notary.notaryd
```
```

**Review by gruyaume — CHANGES_REQUESTED** [review]

File: docs/how-to/hsm.md (hunk 29 above)
Comment (saltiyazan) [reply docs/how-to/hsm.md]: :white_check_mark:

File: docs/how-to/hsm.md (hunk 28 above)
Comment (saltiyazan) [reply docs/how-to/hsm.md]: moved

File: docs/how-to/hsm.md (hunk 27 above)
Comment (saltiyazan) [reply docs/how-to/hsm.md]: :white_check_mark:

File: docs/how-to/hsm.md (hunk 25 above)
Comment (saltiyazan) [reply docs/how-to/hsm.md]: :white_check_mark:

**Review by gruyaume — CHANGES_REQUESTED** [review]
There are still unresolved discussion. Once those are addressed I'm ready to approve the PR.

File: docs/reference/config_file.md (hunk 10 above)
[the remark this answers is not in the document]
Comment (saltiyazan) [reply docs/reference/config_file.md]: We discussed this with @DanielArndt and @kayra1 
We agreed that allowing to omit `encryption_backend` fully will make no backend our default configuration and we shouldn't do that as we should default to the most secure option when possible and definitely shouldn't default to the least secure one, so the user should explicitly configure Notary to have no backend.
We agreed too that having the string "none" feels like using a magic word an is semantically wrong.
I reworked this so now `encryption_backend` is a map and an empty map means no backend
so `encryption_backend: {}`

File: internal/config/config.go [hunk 30]
```diff
@@ -9,6 +9,69 @@ import (
...
+		return fmt.Errorf("encryption_backend must be either 'none' or a map of named backend")
+	}
+
+	var backends map[string]NamedBackendConfigYaml
+	if err := node.Decode(&backends); err != nil {
+		return fmt.Errorf("encryption_backend must be either 'none' or a map of named backends")
+	}
+
+	if len(backends) == 0 {
+		return fmt.Errorf("encryption backend configuration is missing")
```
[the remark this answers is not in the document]
saltiyazan: It is now.

kayra1 (approved):
I think this is a good starting point, so I'm approving it pending @gruyaume's comments are also resolved.

Also a minor comment.

PR #257, opened 2025-07-09
feat: add authorization system

gruyaume:
# Description

Add an authorization system in Notary. The approach is very similar to what we had before but is way more extensible. We can now easily create new roles. 

This task is an implementation of the [TE197 - Role-based access in Notary](https://docs.google.com/document/d/1j3Jq9K4W_Jpouw5wNGqoXUOcHafjBu9H6PTgUqQcsIQ/edit?tab=t.0) specification. 

## Main differences
- We change "Permissions" for "Role ID" in the JWT claim
- We associate each API route with 1 permission
- We create the concept of roles. We start here with 2 roles, that have the same ability has the previously defined "admin" and "user roles". 
- We associate each role with a set of permissions

## Why not use a framework?

In the context of reviewing the spec, @kayra1 mentioned the possibility of using a framework like [casbin](https://casbin.org/). After implementing this change here I'm quite convinced about **not** using a framework. Not only is the change here (almost) code neutral but it's much easier to wrap one's head around. No need to learn new concepts, file formats, etc. everything is in code here.

## What about the other roles

The spec defined more roles than the ones we implemented. Here the goal was to start with the appropriate structure and maintain the same capabilities as before. We will add the remaining roles in a different PR. Adding new roles will simply mean updating the role/permission mapping.

# Checklist:


ui/src/types.ts
```diff
@@ -30,7 +30,7 @@ export type CertificateAuthorityEntry = {
 export type User = {
   exp: number;
   id: number;
-  permissions: number;
+  role_id: number;
```
saltiyazan, line 33: This is very nicely typed in the Go code. Should we do something similar and avoid having magic values like `auth.user?.role_id == 0`?

saltiyazan (changes requested):
Looks good.
I only have two minor comments.

ui/src/types.ts
gruyaume: Agreed, I just found how to do that! Done.

saltiyazan (approved):

kayra1 (approved):

PR #296, opened 2026-02-12
feat: add changing roles as an ability

kayra1:
# Description

This PR lets users with the correct permission modify any other user's role.

# Checklist:


internal/db/db_users.go
```diff
@@ -84,6 +84,16 @@ func (db *DatabaseRepository) UpdateUserPassword(filter UserFilter, password str
 	return UpdateEntity(db, db.stmts.UpdateUser, userRow)
 }
 
+// UpdateUserRole updates the role_id of the given user.
+func (db *DatabaseRepository) UpdateUserRole(filter UserFilter, roleID RoleID) error {
+	if roleID < 0 || roleID > 3 {
```
VercinGor: check logic here, I presume this should be locked to admin only instead of all the roles

internal/server/handlers_accounts.go
```diff
@@ -451,3 +462,62 @@ func ChangeMyPassword(env *HandlerDependencies) http.HandlerFunc {
...
+			writeResponse(w, http.StatusInternalServerError, "", nil, env.SystemLogger)
+			return
+		}
+		if idNum == 1 {
+			err = errors.New("updating the default admin account role is not allowed")
+			writeResponse(w, http.StatusBadRequest, "updating the default admin account role is not allowed.", err, env.SystemLogger)
+			return
+		}
+
+		claims, err := getClaimsFromCookie(r, env.Database.JWTSecret, env.AuthnRepository)
```
VercinGor: would move this before the user fetch, so unauthorized users won't trigger anything else.

VercinGor (approved):
lgtm, just review the couple of comments, thanks.

internal/server/handlers_accounts.go
kayra1: agreed

internal/db/db_users.go
kayra1: good catch, but this is actually an unnecessary error check if the roleID doesn't exist. just removed it completely.

VercinGor: I see I got this wrong firstly on what it was intended for, but seems redundant agreed.

VercinGor (approved):

PR #330, opened 2026-04-07
feat: re-enable onboarding flow

kayra1:
# Description

To shrink the size of the OIDC changes and to rethink our approach to onboarding, I had simplified the flow to include a default admin account. This PR reverts that change and brings back the original mechanism of the first account automatically being an admin. This should be much better than the simple version and it's tried and tested from before.

# Checklist:


internal/server/handlers_oidc.go
```diff
@@ -140,19 +140,35 @@ func CallbackOIDC(env *HandlerDependencies) http.HandlerFunc {
...
 
-			// Auto-provision new OIDC user with RoleReadOnly (ID=3)
+			// Auto-provision new OIDC user
 			// Email is optional - the OIDC subject is the primary identifier
 			emailOrPlaceholder := email
 			if emailOrPlaceholder == "" {
 				emailOrPlaceholder = "(none)"
 			}
 
+			// Determine role: first user gets admin, subsequent users get read-only
```
saltiyazan, line 158: Are all oidc users read-only? can the role be changed?

internal/server/middleware.go
```diff
@@ -279,6 +279,26 @@ func requirePermission(
 	}
 }
 
+// firstUserOrAdmin allows unauthenticated access when zero users exist (first-run setup).
+// This enables the initial admin account to be created without pre-existing credentials.
+// Once any user exists, it falls back to requirePermission with adminOnly access.
+func firstUserOrAdmin(env *HandlerDependencies, handler http.HandlerFunc) http.HandlerFunc {
```
saltiyazan, line 285: Should we set a lock on the first user so we don't have a race condition where two first users can be created at init time if two requests are sent simultaneously? or is that too much of an overkill for that edge case?

internal/server/handlers_accounts.go
```diff
@@ -270,9 +281,14 @@ func DeleteAccount(env *HandlerDependencies) http.HandlerFunc {
...
 		}
-		if account.RoleID == db.RoleID(RoleAdmin) {
-			err = errors.New("deleting an Admin account is not allowed")
-			writeError(w, http.StatusBadRequest, "deleting an Admin account is not allowed.", err, env.SystemLogger)
+		numUsers, err := env.Database.NumUsers()
+		if err != nil {
+			writeError(w, http.StatusInternalServerError, "Internal Error", err, env.SystemLogger)
+			return
+		}
+		if numUsers <= 1 && env.AuthnRepository == nil {
```
saltiyazan, line 289: Wouldn't this allow for the admin user be deleted while no other admin users exist? leaving the system with no admin users?

saltiyazan (changes requested):
Looks good, I have a few questions.

internal/server/handlers_oidc.go
kayra1: yes, any new OIDC user is automatically assigned read-only, and an admin can change their role later.

internal/server/middleware.go
kayra1: I would consider it overkill. This is something that only the person who's first deploying notary will do once, and they'd have to do both calls at the same millisecond.

internal/server/handlers_accounts.go
kayra1: yes. i think in some situations that may be preferable. having no superusers means there's no way for anyone to modify the users of the system anymore, making it easier to secure. Just like how Vault prefers users to delete the root token after the first user is created through an auth method.

saltiyazan: I see, but my problem is that if we delete the admin user without having another admin user we would get into a kind of a deadlock, no?

kayra1: yes, that is on purpose. that is also possible on vault.

saltiyazan (approved):
Looks good. Thanks!

PR #350, opened 2026-05-05
feat: modernize UI

kayra1:
# Description

This PR changes a bunch of fundamental things about the frontend.

## Use Bun over Nodejs

Bun is an extremely fast javascript runtime that comes with a nice CLI that includes a ton of quality of life features that help with building and distributing the frontend and increase the speed of the CLI. 

## Use Biome over eslint

Biome is a new toolchain for typescript that has a bunch of improvements and quality of life features on top of being fast. It includes a linter, a formatter, an extra assister that comes with a bunch of useful automations like sorting imports, and is also an LSP.

## Use vite over next

This is necessary for the next step where we replace nextjs with a frontend only router framework. Vite is a typescript bundler/compiler that converts react + typescript into browser-native javascript. It is extremely fast, has a bunch of helpful developer tools to facilitate dev mode, and is an all around great option

## Use tanstack router over nextjs

We used a full stack framework, but at the end of the day we are embedding a frontend into a go backend, which means we aren't using any of the features nextjs has but introducing all of the build pains and bloat. React routers are explicitly frontend only and are designed to be static exports, and the snappiness of the UI introduced with this PR is proof that this is valuable on top of getting rid of a lot of bloat. This gets rid of a bunch of issues, including unnecessary page refreshes. I chose tanstack router over other options because it has extra integrations with tanstack query, which we already use. It's a stable router that's SOTA but also supported long term and we already depend on Tanstack Query which is from the same company.

# Checklist:


kayra1:
never doing the agent thing again.

saltiyazan (changes requested):
Looks good to me. Nice work. Please rebase on main and make sure the ACME servers page is visible.

saltiyazan (approved):

PR #377, opened 2026-08-31
feat: Use dqlite instead of sqlite in Notary

saltiyazan:
# Description

## Summary

This PR implements **HA stage 1** by moving Notary's application database from a local SQLite file to a **single-node dqlite cluster**.

This is an intentional breaking change for the HA MVP. Existing SQLite database files are not upgraded in place. There is no cluster join or membership management yet.

## Architecture

Notary continues to use sqlair over `*sql.DB`. The implementation uses `github.com/canonical/go-dqlite/v3/app` directly because microcluster only exposes transaction callbacks, which cannot provide the `*sql.DB` required by sqlair and OpenFGA.

`db_path` now identifies a directory rather than a SQLite file. It contains:

- dqlite identity, raft, and application database files
- `openfga.sqlite`, a separate SQLite sidecar used by OpenFGA

An empty directory bootstraps a one-node cluster. A directory containing `info.yaml` resumes the existing node.

The optional `cluster.address` setting controls the dqlite bind address and defaults to `127.0.0.1:9000`.

## Schema management

Notary schema updates now run automatically during `notary start`.

The previous Goose migrations have been replaced by ordered schema updates recorded in a `schema_updates` table. Consequently:

- `notary migrate` has been removed
- `notary start -m` has been removed
- Goose remains only for OpenFGA's own sidecar schema

Field-level AES-256-GCM encryption and the existing PKCS#11, Vault, and unwrapped KEK options are unchanged. Dqlite itself does not provide encryption at rest.

## OpenFGA

OpenFGA's SQLite storage driver is not compatible with dqlite, so authorization data is stored in `{db_path}/openfga.sqlite`.

The sidecar uses OpenFGA's standard SQLite DSN settings, including WAL mode, busy timeout, and immediate transaction locking. All other Notary application data remains in dqlite.

This will be addressed in the upcoming phases

## Backup and restore

The existing commands remain:

- `notary backup`
- `notary restore`

They now operate on a `tar.gz` archive of the complete data directory, including both dqlite and OpenFGA data.

The Notary daemon must be stopped before either operation. Copying dqlite raft files while the node is writing is unsafe.

Backup behavior:

- `-d` identifies the data directory
- `-f` must include a destination directory
- the command creates a unique timestamped archive
- destinations inside the data directory, including symlink aliases, are rejected

Restore extracts and validates the archive before replacing the data directory. After restoration, `cluster.address` must match the address stored in the restored node's `info.yaml`.

A future stage can use dqlite's live `Dump()` API to remove the shutdown requirement for backups. Restore will still require the daemon to be stopped.

## Packaging and CI

The snap, rock, unit-test workflow, and TiCS workflow now install and link against libdqlite with CGO enabled.

The rock smoke test no longer passes the removed `-m` option.

## Compatibility

This PR does not provide an in-place migration from an existing SQLite database. Pointing the new version at an old SQLite file will fail because `db_path` is now expected to be a directory.

## Out of scope

This PR does not add:

- additional cluster members or multi-node raft
- join tokens or membership APIs/CLI/UI
- leader-only ACME processing
- live dqlite backups
- automatic migration from the previous SQLite store

## Suggested review areas

The most important areas to review are:

- dqlite bootstrap, resume, and shutdown lifecycle
- automatic schema-update ordering and atomicity
- OpenFGA sidecar initialization
- backup path safety and restore behavior
- snap/rock libdqlite packaging

## Test plan

```bash
export CGO_ENABLED=1

# If libdqlite is installed outside the system paths:
export CGO_CFLAGS="-I$HOME/.local/dqlite/include"
export CGO_LDFLAGS="-L$HOME/.local/dqlite/lib"
export LD_LIBRARY_PATH="$HOME/.local/dqlite/lib:$LD_LIBRARY_PATH"

go test ./...
make notary
```

# Checklist:


cmd/migrate.go
kayra1, line 1: Alessandro specifically asked me to add the goose migrations to notary, please talk to him first about removing it. It could be good to even get his review.

internal/cluster/node.go
```diff
@@ -0,0 +1,105 @@
...
+	if opts.Dir == "" {
+		return nil, errors.New("database directory is required")
+	}
+	if err := os.MkdirAll(opts.Dir, 0o700); err != nil {
+		return nil, fmt.Errorf("create database directory: %w", err)
+	}
+
+	var appOpts []app.Option
+	if opts.Address != "" {
+		appOpts = append(appOpts, app.WithAddress(opts.Address))
```
kayra1, line 56: Does this mean the cluster has no TLS?

kayra1 (commented):
Although I haven't looked at this in a while, I'm mostly only worried about getting rid of the existing migration stuff for the custom schema generation.

internal/cluster/node.go
saltiyazan: No at this point not yet, those are coming in the next PR (in this one it's a one node cluster)

cmd/migrate.go
saltiyazan: Brought it back.

VercinGor (approved):
Looks good, played around with it locally, all seems in order didn't bump into any hurdles so far.

PR #380, opened 2026-09-03
feat: Drop OpenFGA in favor of a role id in the database

saltiyazan:
# Description

OpenFGA in Notary was only used for a few roles on one system object, not a real relationship graph, and it stored that in its own SQLite file next to the app. dqlite can replicate Notary’s database, but not that sidecar, and OpenFGA’s SQLite driver cannot sit on dqlite, so a cluster would have split, inconsistent permissions. For that simple role model it was enough to read users.role_id from the same dqlite database the rest of Notary already uses, keep the same Check API, and drop OpenFGA rather than run a second store or rewrite OpenFGA for raft.

# Checklist:


VercinGor (approved):
Looks good, local testing didn't surface any issues. Thanks

PR #381, opened 2026-09-04
feat: Implement HA clustering features.

saltiyazan:
# Description

This branch adds **multi-node clustering** on top of the dqlite work already on `main`.

**What an operator does**

1. Start the first node (empty data directory, no `--join`).
2. On that node: `notary cluster add node2` — prints a token.
3. On the new machine: `notary start --join '<token>'`.
4. `notary cluster list` / `notary cluster remove` as needed.

The HTTP API matches that (`list` / `add` / `remove`). There is no cluster page in the UI.

**How trust works (plain language)**

The token is a **short-lived, one-use pass**, not the cluster private key. The new node uses it to talk to an existing node over HTTPS, checks it is talking to the right server, and only then receives the cluster certificate. After that, dqlite can join.

You do **not** copy `cluster.crt` / `cluster.key` to the joiner. The first node can create those files itself if you omit `cluster.tls`.

**Also included**

- Backup will not run while Notary is using the data directory.
- Logging does not have to be in the config file.
- The cluster database migration is numbered `00004` (HA-02 already used `00003`).

**Not in this PR**

ACME-only-on-the-leader, per-node raft certificates, or a cluster UI.

# Checklist:


VercinGor (commented):
One potential issue that I bumped into that need a bit attention to verify or discard. 

In `/internal/backends/authentication/jwt.go` there is a function that `SetUpJWTSecret`.

When a token does not exist it generates a new token and stores it into the database, but in that case it does not end up stored into memory of the working process and it exits. This is no issue in single node operation, but with this addition of clustering it surfaces the problem when another node checks in memory and gets nil, while the database check passes. 

adding the missing `database.JWTSecret = jwtSecret` in the conditioning block seems to be the way  forward so both paths are successful.

VercinGor (approved):
Validated most cases locally, all seems to be in order except the one case I commented about. Once that is solved/discarded this is good to go.

saltiyazan:
> One potential issue that I bumped into that need a bit attention to verify or discard.
> 
> In `/internal/backends/authentication/jwt.go` there is a function that `SetUpJWTSecret`.
> 
> When a token does not exist it generates a new token and stores it into the database, but in that case it does not end up stored into memory of the working process and it exits. This is no issue in single node operation, but with this addition of clustering it surfaces the problem when another node checks in memory and gets nil, while the database check passes.
> 
> adding the missing `database.JWTSecret = jwtSecret` in the conditioning block seems to be the way forward so both paths are successful.

Fixed in f36023b7c877ea620e664e37ddf31ff9539d8c6b

PR #382, opened 2026-09-07
feat: Complete dqlite HA operator path and CA-mode cluster TLS

saltiyazan:
# Description

This branch sits on top of dqlite and LXD-shaped join already on main. It is the rest of the operator path for a real multi-node Notary, plus an optional CA-mode for cluster TLS.

ACME signing is leader-only. Followers return 409 with the leader’s HTTPS address so clients retry on the right node instead of racing the same public CA order. Internal CA signing is unchanged and can still hit any member. The 409 is a gate, not a lock; do not put ACME sign behind a VIP that hides which unit you reached.

The HA operator path records each member’s `api_address`, keeps OIDC state in dqlite, and serializes goose migrations with a cluster-wide lock. Join tokens stay one-time tickets: redeem is unauthenticated `POST /api/v1/cluster/join`, consume is one DELETE, and a failed membership listing restores the token so the same ticket can be retried. If redeem already succeeded and dqlite join then fails, the token is spent. `notary cluster recover` inspects raft term/index and, with `--force`, rewrites the node to a single voter. That is destructive and documented that way. The Cluster UI lists members and can mint or remove; backup/restore runbooks treat `db_path` as cluster credential. Metrics stay cluster-wide: scrape one node.

The snap configure hook is fixed so clustered deploys can wire the same settings.

Default cluster TLS is still the shared self-signed pair in `db_path`. CA mode is opt-in and all-or-nothing:

```yaml
cluster:
  tls:
    ca_path: /etc/notary/cluster-ca.crt
    cert_path: /etc/notary/unit.crt
    key_path: /etc/notary/unit.key
    peer_san: "notary-cluster"
```

Notary does not mint or revoke that CA. Each unit presents its own leaf; trust is the CA plus `peer_san` (DNS or URI) on every peer. Host/IP SANs are only for dial `ServerName` from the raft address. CA mode never writes `cluster.key`, never generates a shared pair on resume, and join redeem returns addresses only—no private key, no CA in the body. Shared and CA members cannot mix. `peer_san` is checked with `VerifyConnection` so it still runs on TLS session resumption. dqlite has no protocol auth: TLS trust on the dqlite port is authorisation. Join tokens only constrain `notary start --join`. After `cluster remove`, stop the unit and, in CA mode, revoke or expire its leaf.

Not in this branch, on purpose: dqlite role management (go-dqlite still promotes voters), HTTPS raft, live `Dump()`, an ACME claim row, custom roles, and raft address rewrite.

# Checklist:


ui/src/routes/cluster/-components/asideForm.tsx
```diff
@@ -0,0 +1,144 @@
+import {
```
VercinGor, line 1: Missing padding in side form.

ui/src/routes/cluster/-components/table.tsx
```diff
@@ -0,0 +1,108 @@
...
+					onClick={() => {
+						setAsideOpen(true);
+					}}
+				>
+					Add member
+				</Button>
+			}
+		>
+			<p>
+				Members share one dqlite database. Add a member to mint a join token,
```
VercinGor: suggestion change for better clarity of actin: "All members share a single dqlite database. Click **Add member** to generate a join token, then start your new node with an empty data directory using notary start --join"

ui/src/routes/cluster/-components/asideForm.tsx
```diff
@@ -0,0 +1,144 @@
...
+			controls={
+				<Button onClick={handleClose} hasIcon>
+					<i className="p-icon--close" />
+				</Button>
+			}
+		>
+			{token ? (
+				<>
+					<p>
+						Start the new node with an empty <code>db_path</code> and this
```
VercinGor: suggestion - Start the new node with an empty db_path and this token. Until redeemed or expired, anyone with the token can access the cluster private key.

VercinGor (approved):
Local testing seems ok, didn't manage to brake anything and works as expected.

left few remarks on the UI, main one the side form padding to match the rest of the dashboard. Suitable icon as well for the `Cluster` item on the side menu. 

The rest are minor wording suggestions. We can discuss the final UX/UI in the UI rework effort.

ui/src/routes/cluster/-components/asideForm.tsx
saltiyazan: Fixed

ui/src/routes/cluster/-components/table.tsx
saltiyazan: fixed

ui/src/routes/cluster/-components/asideForm.tsx
saltiyazan: suggestion applied

PR #384, opened 2026-09-10
fix: make cluster join-token lifecycle retryable

saltiyazan:
# Description

fix: make cluster join-token lifecycle retryable

A node joining with a one-time token could get permanently stuck in
three ways:

- When redeem succeeded but listing members failed (typically a
  leadership change expiring the 15s request deadline), the API told
  the client to retry with the same token, but the token restore ran
  on that same expired context and silently failed. The restore now
  gets its own 10s budget via context.WithoutCancel.

- When the dqlite join then failed (e.g. unreachable join addresses),
  app.New had already written info.yaml, so every retry — even with a
  fresh token — was rejected with "join token is only used the first
  time this node starts". Start now snapshots whether the directory
  had prior state and removes the identity files (info.yaml,
  cluster.yaml, join) after a failed first start. Directories with
  pre-existing state are never cleaned, and neither is the failure
  path after the node was ready.

- A join token carried only external_hostname as its redeem address,
  so a hostname that doesn't resolve from the joiner's network (e.g.
  an internal-only FQDN) made redeeming impossible. Tokens now also
  carry the cluster bind address as a fallback; joiners try addresses
  in order with a per-address dial timeout (redeem stays fingerprint-
  pinned, not hostname-pinned). Issuance warns when the primary
  address is unreachable from the issuer — a heuristic, since
  issuer-side checks can't prove joiner-side reachability.

Also adds coverage for the wrong-secret redeem branch, confirming a
wrong guess neither succeeds nor burns the ticket.

# Checklist:


VercinGor (approved):
Looks in order, no regression in local validating.

